PR #395, opened 2014-07-23
Initial pass at Camel messaging implementation

danveloper:


alkemist:
@danveloper I'm going to close this off. Sorry we never got it integrated. We've got the code if we revisit in the future.

alvarosanchez:
So any plan to revisit the integration of this feature? I think that adding Camel support would be a great contribution to Ratpack, as it will open many doors to use the supported camel components.

alkemist:
None in the short term. It's something we'd like to do one day though.

PR #523, opened 2014-12-14
Added basic request logger implementation

kyleboon:
Not sure this is the best place for this code, but creating a new module seemed like overkill.


kyleboon:
I updated the UUID generator to be non blocking and using the fastest one I've found so far. I will want to add a few more tests for this, I should have time tomorrow to get that done.

alkemist:
We are still in flight here right?

johnrengelman:
Rebased and merged as https://github.com/ratpack/ratpack/commit/93151f94235684707f24429d03ac2c7a11f33018

Issue #538, opened 2015-01-03
ratpack-pac4j authentication handler does not properly handle ajax requests.


jeff-blaisdell:
While requesting a path that requires authentication the Pac4jAuthenticationHandler passes a hard coded 'false' to the client's redirection method for the ajaxRequest argument.  This causes ajax requests to be sent redirects rather than the expected 401 response.

The problem method:
```
  private void initiateAuthentication(final Context context) {
    final Request request = context.getRequest();
    request.get(SessionStorage.class).put(SAVED_URI, request.getUri());
    final Clients clients = request.get(Clients.class);
    final RatpackWebContext webContext = new RatpackWebContext(context);
    context.blocking(() -> {
      clients.findClient(name).redirect(webContext, true, false);
      return null;
    }).onError(ex -> {
      if (ex instanceof RequiresHttpAction) {
        webContext.sendResponse((RequiresHttpAction) ex);
      } else {
        throw new TechnicalException("Failed to redirect", ex);
      }
    }).then(ignored -> webContext.sendResponse());
  }
```

The specific line is:
```
      clients.findClient(name).redirect(webContext, true, false);
```

davidmc24:
Good point.  Given that context, hard-coding `false` clearly isn't the right thing to do.  Do you have any suggestions on how this should be handled?  That is, how should we determine whether to pass `true` vs. `false`?  Perhaps a new method on `Authorizer`, `boolean isAjaxRequest(Context)`, with a default of `false`?

jeff-blaisdell:
Above is the PR I started on.

Sniffing for a X-Requested-With header set to XMLHttpRequest I think is a fairly common pattern.  I believe it was started with jQuery.  I could also see the benefit of making something a little fancier than I did in my PR - something like you said - another method on authorizer would work as well.  

I personally have never had a reason to do anything other then the X-Requested-With sniffing, but I'm sure someone could come up with a case.

davidmc24:
Thanks for the pull request!  I like the approach of using the `X-Requested-With` header better than a default of always `false`.  I think you're right that it will probably handle most cases, but I'm not quite comfortable with hardcoding that behavior either.

What do you think about making it a function in Pac4jAuthenticationHandler, with a setter in case someone wants to override it?  It's an internal class, but I think that's a reasonable compromise for now. 

Maybe something like
[...]

PR #539, opened 2015-01-03
Pac4j authentication handler should test for AJAX request before redirection.

jeff-blaisdell:
Initial attempt to fix https://github.com/ratpack/ratpack/issues/538

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/Request.java
+++ b/ratpack-core/src/main/java/ratpack/http/Request.java
@@ -130,6 +130,12 @@ public interface Request extends MutableRegistry {
   HostAndPort getLocalAddress();
 
   /**
+   * A flag representing whether or not the request originated via AJAX.
+   * @return A flag representing whether or not the request originated via AJAX.
+   */
+  boolean isAjaxRequest();
+
+  /**
    * {@inheritDoc}
    */
   @Override
--- a/ratpack-core/src/main/java/ratpack/http/internal/DefaultRequest.java
+++ b/ratpack-core/src/main/java/ratpack/http/internal/DefaultRequest.java
@@ -185,6 +185,10 @@ public class DefaultRequest implements Request {
     }
   }
 
+  public boolean isAjaxRequest() {
+    return HttpHeaderConstants.XML_HTTP_REQUEST.equalsIgnoreCase(headers.get(HttpHeaderConstants.X_REQUESTED_WITH));
+  }
+
   @Override
   public TypedData getBody() {
     if (body == null) {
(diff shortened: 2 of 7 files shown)


ratpack-pac4j/src/main/java/ratpack/pac4j/internal/Pac4jAuthenticationHandler.java
```diff
@@ -76,4 +78,9 @@ private void initiateAuthentication(final Context context) {
       }
     }).then(ignored -> webContext.sendResponse());
   }
+
+  private boolean isAjaxRequest(Request request) {
+    Headers headers = request.getHeaders();
+    return "XMLHttpRequest".equals(headers.get(HttpHeaderConstants.X_REQUESTED_WITH));
```
davidmc24: Might be worthwhile to move `"XMLHttpRequest"` into `HttpHeaderConstants` as well, actually.

johnrengelman: Probably is worthwhile to have this method defined on the `Request` API itself, don't you think? I can image there are a number of place where you'd like to know if a request is ajax or not.

davidmc24: Yeah, you're probably right.  I'm always hesitant to make hard use of non-standard HTTP headers, but this one is probably a case where the value is worth it.

jeff-blaisdell: Good points.  I'll update.

jeff-blaisdell: Above changes have been made.

ratpack-core/src/main/java/ratpack/http/Request.java
```diff
@@ -130,6 +130,12 @@
   HostAndPort getLocalAddress();
 
   /**
+   * A flag representing whether or not the request originated via AJAX.
+   * @return A flag representing whether or not the request originated via AJAX.
+   */
+  Boolean isAjaxRequest();
```
davidmc24: If it's not meaningful for this to be `null`, might want to change this to return `boolean`.

ratpack-core/src/main/java/ratpack/http/internal/DefaultRequest.java
```diff
@@ -185,6 +185,10 @@ public String oneCookie(String name) {
     }
   }
 
+  public Boolean isAjaxRequest() {
+    return HttpHeaderConstants.XML_HTTP_REQUEST.equals(headers.get(HttpHeaderConstants.X_REQUESTED_WITH));
```
davidmc24: Should this match be case-sensitive, or case-insensitive?

davidmc24:
Pretty close.  I think with the next round of changes this will be ready to merge.  Please add yourself to the list of contributors in `/ratpack-manual/src/content/chapters/99-about-the-project.md`.

jeff-blaisdell:
Made changes.

ratpack-core/src/main/java/ratpack/http/internal/DefaultRequest.java
jeff-blaisdell: Case insensitive would probably be less of a hassle for people.

davidmc24:
Looks good to me.  Thanks again!

PR #542, opened 2015-01-05
Adding benchmarking sub project

### LINKED ISSUE #531: Make it easy to write temporary micro benchmarks in the build
I'd love to have some permanent stuff set up in the build to make it effortless to write new JMH micro benches. Something like another subproject in the build set up so we can just drop new benchmarks in for experimenting.

Once we have this set up, we might set up some permanent benchmarks and track them over time on http://perf.ratpack.io.

kyleboon:
Closes #531 

There are 2 simple examples of benchmarks included. I'm happy to write some tests for specific portions of ratpack as well for some better examples but I wanted to make a PR because I think @danhyun wants to write some benchmarks.


danhyun:
Thanks so much @kyleboon! I'll take a look at the examples tonight.

kyleboon:
@danhyun this has some more complicated examples: https://github.com/nitsanw/jmh-samples including parameterized tests and shared state between threads

ratpack.gradle
```diff
@@ -123,10 +123,10 @@ ext {
     "ratpack-core", "ratpack-groovy", "ratpack-guice", "ratpack-session", "ratpack-test", "ratpack-groovy-test",
     "ratpack-manual", "ratpack-gradle", "ratpack-handlebars", "ratpack-remote", "ratpack-remote-test", "ratpack-jackson",
     "ratpack-codahale-metrics", "ratpack-h2", "ratpack-thymeleaf", "ratpack-rx", "ratpack-hikari", "ratpack-newrelic",
-    "ratpack-pac4j", "ratpack-hystrix", "ratpack-spring-boot", "ratpack-config"
+    "ratpack-pac4j", "ratpack-hystrix", "ratpack-spring-boot", "ratpack-config", "ratpack-benchmark"
```
johnrengelman: I don't think we'll want to publish anything from this subproject, so this can be removed. Which means it can be removed below as well.

alkemist:
Completely forgot about this, sorry.

Merging…

PR #543, opened 2015-01-05
Added a simple example in Javadoc on how to use HttpClient

gsaqui:
Adding javadoc example on how to use HttpClient for a simple GET and POST request


beckje01:
Merged by hand Thanks for the docs.

alkemist:
Thanks @gsaqui !

PR #557, opened 2015-01-16
Update reactive-streams to 1.0.0.RC1

smaldini:


davidmc24:
Thanks for the pull request.  It looks like there were changes to the reactive-streams spec between these versions that will require code changes to remain compliant.  With just updating the version of the reactive-streams library, we have failing tests.

Would you care to take a shot at making the updates?

johnrengelman:
This was done in https://github.com/ratpack/ratpack/commit/3355db190a35defb5e00ef45a577915b3525d9f3

PR #559, opened 2015-01-21
Updated Heroku documentation 

jkutner:
Removing discussion on setting jdk version. The default on Heroku is JDK8 now, so it is probably best just to remove this section.


davidmc24:
Makes sense to me.  Thanks for the contribution.

Issue #551, opened 2015-01-14
Add support for Slfj MDC support


alkemist:
See MDC section of http://www.slf4j.org/manual.html

For us to support this, we'd need to manage swapping out thread bound MDCs as executions acquire a thread. This could be done quite easily with an ExecInterceptor implementation.

Somewhat dependent on #550

davidmc24:
See also: http://www.slf4j.org/api/org/slf4j/MDC.html

This interceptor should:
* Maintain a `ConcurrentMap<Execution, Map<String, String>>`, where each value is a MDC context map
* Before running the continuation, check the internal map, and if an entry for the current execution is available, set it in the MDC; if not, clear the MDC
* Run the continuation
* After running the continuation, get the context map from the MDC.  If there was one, store it in the internal map.

Together, this means that, once the interceptor is registered, the user can directly use the SLF4J MDC API, and it will just work.

alkemist:
Actually, `Execution` is a registry. We could just put the context map in there which removes the need for managing a concurrent map.

zedar:
I'm about to finish implementation of MDCInterceptor, but I've got 2 questions:
1. I understand that the goal is to pass MDC entries between threads in the same execution. Then blocking, promise with execController.control.exec() should inherit MDC entries. But for example org.slf4j.helpers.BasicMDCAdapter uses InheritableThreadLocal, so entries are inherited between child threads by default. Log4j uses ThreadLocal when special property (isThreadContextMapInheritable) is not set else it uses InheritableThreadLocal too. If logging providers have similiar functionality, do we need this interceptor anyway?

2. In order to unit test MDC functionality, we need one of the logging providers on class path. Without it Slf4j uses NOPMDCAdapter. Slf4j tries to get MDCAdapter via org.slf4j.impl.StaticMDCBinder. Could you give me some hints? Should I add logging provider dependency for ratpack-core tests or should I provide custom stubs for StaticMDCBinder and MDCAdapter?

davidmc24:
I think the key things that are different about Ratpack that require a different approach to handling the MDC are (please correct me if I'm wrong, @alkemist):

* The execution segments for an execution may be executed on unrelated threads in a pool (not child threads)
* On a given thread, you can't assume that the only work performed will be for a single execution until its completion

[...]

Issue #415, opened 2014-08-08
Run health checks on a non request thread


rhart:
`HealthCheckHandler` should run the health checks on a non request thread so they can block.

PR #588, opened 2015-02-25
Slf4j MDC support

zedar:
Added ```ratpack.logging.MDCInterceptor```.
 * Before running the continuation, it checks if internal map contains any entry.
 * If yes, it puts all of them from internal map to MDC (bounded to current thread).
 * If internal map is empty, it clears MDC.
 * After continuation it gets the entries from MDC's context map and stores them in execution's internal map.

Added small fix to ```ratpack.handlingChain``` Javadoc to force formatting of floating number separator (important for non english operating system).


zedar:
Issue #551

alkemist:
Looking good @zedar 

Left some comments inline.

zedar:
Thanks @alkemist. Pushed fixes after your hints.

ratpack-core/src/main/java/ratpack/logging/MDCInterceptor.java
```diff
@@ -0,0 +1,119 @@
...
+import ratpack.exec.ExecInterceptor;
+import ratpack.exec.Execution;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.slf4j.MDC;
+
+/**
+ * Intercept execution and add support for <a href="http://www.slf4j.org/api/org/slf4j/MDC.html">Slf4j MDC</a> described in the <a href="http://www.slf4j.org/manual.html#mdc">manual</a>.
```
davidmc24: Nitpick, but I believe the canonical capitalization is "SLF4J", not "Slf4j"

ratpack-core/src/main/java/ratpack/logging/MDCInterceptor.java
```diff
@@ -0,0 +1,119 @@
...
+ * Mapped Diagnostic Context (MDC) is a map of key-value pairs provided by client code and then automatically inserted in log messages.
+ * The underlying logging framework has to support MDC logging.
+ * Example {@code log4j2} configuration that assumes {@code value} as MDC context map value:
+ * {@code
+ * <Console name="Console" target="SYSTEM_OUT">
+ *   <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg - [%X{value}] %n"/>
+ *  </Console>
+ * }
+ * <p>
+ * When interceptor is registered with {@link ratpack.exec.ExecControl#addInterceptor}, it is possible to directly use Slf4j MDC API.
```
davidmc24: Capitalization of Slf4j here too.

ratpack-core/src/main/java/ratpack/logging/MDCInterceptor.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+public class MDCInterceptor implements ExecInterceptor {
+
+  public static class MDCMap extends HashMap<String, String> {
+  }
+
+  private final Execution current;
+
+  public MDCInterceptor() {
+    current = ExecControl.current().getExecution();
```
davidmc24: Is there a reason to do this initialization upon construction?  I would think it would make sense to do it all in `intercept`.  With the current implementation, I'm worried that it may get one execution in the constructor, and then have `intercept` called multiple times from within different executions and behave incorrectly.

zedar: ```ExecControl.current().getExecution();``` called from inside ```MDCInterceptor.intercept()``` throws ```ratpack.exec.ExecutionException: Current thread (ratpack-blocking-3-1) has no bound execution``` when intercepts ```blocking``` section ```//3```.

    get("foo1") {
        // 1
        MDC.put("value", "foo1")
        values << MDC.get("value")
        blocking {
          // 3
          values << MDC.get("value")
          MDC.put("value", "foo2")
          values << MDC.get("value")
        }.then {
          // 4
          values << MDC.get("value")
          render "foo1"
        }

I assume that only forked executions should start with clear MDC context. For example ```//3``` should have the same MDC context as ```//2``` and ```//3``` in above example. ```Execution``` is used only as ```Registry``` for ```MDCMap```.

alkemist: I think we should change the ExecInterceptor signature to take the Execution as an arg.

alkemist:
Many thanks for the contribution @zedar !

I made some tweaks to the impl and docs: http://www.ratpack.io/manual/0.9.15/api/ratpack/logging/MDCInterceptor.html

zedar:
Thanks @alkemist. Ratpack's execution flow is awesome. Could you suggest another issue to work with?

alkemist:
What about adding support for non-blocking health checks? #415

zedar:
@alkemist: started to work on #415.

alkemist:
Great :)

PR #605, opened 2015-03-09
metrics regex reporter filter

gschrader:
It would be nice to have the test also check the jmx reporter, the other tests aren't checking it currently so I'll see how hard it would be to add that.


alkemist:
Hi @gschrader 

Looks like there are some build failures with this change.

gschrader:
oops, sorry I will fix it

rhart:
Thanks @gschrader, this is exactly what I was thinking.

@alkemist is the approach ok to you? Exposing the configurable options of the reporters on our config objects as opposed to just exposing the reporter builders?

alkemist:
@rhart seems like a common bit of config, so +1 from me.

alkemist:
didn't mean to close.

PR #613, opened 2015-03-25
Add support for nonblocking health checks

zedar:
Implementation of #415 issue.

Non-blocking health checks are part of ratpack-core.
Health check returns promise executed by HealthCheckHandler.
It's possible to run all registered health checks or single health check with the given name.
concurrencyLevel determines if health checks run in parallel, in sequence or in group of parallel promises.


ratpack-core/src/main/java/ratpack/health/HealthCheck.java
```diff
@@ -0,0 +1,182 @@
...
+    private final Throwable error;
+
+    /**
+     * Create health check result.
+     *
+     * @param healthy if health check succeeded should be true
+     * @param message message generated by health check
+     * @param error exception thrown by health check execution
+     */
+    public Result(boolean healthy, String message, Throwable error) {
```
alkemist: We should make this constructor private and remove the javadoc.

ratpack-core/src/main/java/ratpack/health/HealthCheck.java
```diff
@@ -0,0 +1,182 @@
...
+    public String getMessage() {
+      return message;
+    }
+
+    /**
+     * Get exception thrown by health check.
+     *
+     * @return exception thrown by unhealthy application component
+     */
+    public Throwable getError() {
```
alkemist: Needs @Nullable

ratpack-core/src/main/java/ratpack/health/HealthCheck.java
```diff
@@ -0,0 +1,182 @@
...
+    public boolean isHealthy() {
+      return healthy;
+    }
+
+    /**
+     * Get message generated by health check.
+     *
+     * @return additional message either for healthy or unhealthy application component
+     */
+    public String getMessage() {
```
alkemist: Needs @Nullable

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Handler that runs and renders health checks executing in non-blocking mode
+ * <p>
+ * This handler queries {@code context} for either all {@link ratpack.health.HealthCheck}s or health check with th given name.
```
alkemist: typo: `th given name`

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+ * Then handler gets {@link ratpack.exec.Promise} returning {@link ratpack.health.HealthCheck.Result} for every health check to execute.
+ * {@code concurrencyLevel} determines if promises are executed:
+ * <ul>
+ *   <li>in parallel (if {@code concurrencyLevel=0})</li>
+ *   <li>in sequence (if {@code concurrencyLevel=1})</li>
+ *   <li>in groups (if {@code concurrencyLevel >= 2}). Groups are executed in sequence while promises inside group in parallel</li>
+ * </ul>
+ * <p>
+ * This handler should be bound to an application path, and most likely only for the GET method.
+ * <pre class="java-chain-dsl">
```
alkemist: I wouldn't bother with this example. The one below is enough.

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+ *         f.success(HealthCheck.Result.healthy());
+ *       });
+ *     }
+ *   }
+ *
+ *   public static void main(String... args) throws Exception {
+ *     EmbeddedApp.of(s -> s
+ *       .registry(Guice.registry(b -> b
+ *         .bind(FooHealthCheck.class)
+ *         .bindInstance(HealthCheckResultsRenderer.class, new HealthCheckResultsRenderer())
```
alkemist: We should always add this guy to the base registry (like the file renderer), so users don't need to.

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+ *         .bindInstance(HealthCheckHandler.class, new HealthCheckHandler())
```
alkemist: I'd remove this from the example and just add a directly created instance below.

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+ *       ))
+ *       .handler(HealthCheckHandler.class)
+ *     ).test(httpClient -> {
+ *       assertEquals("foo : HEALTHY", httpClient.getText());
+ *     });
+ *
+ *     EmbeddedApp.of(s -> s
```
alkemist: I wouldn't bother with this example.

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+ * }</pre>
+ *
+ * @see ratpack.health.HealthCheck
+ * @see ratpack.health.HealthCheckResults
+ * @see ratpack.health.HealthCheckResultsRenderer
+ */
+public class HealthCheckHandler implements Handler {
+  public static final String DEFAULT_NAME_TOKEN = "DEFAULT";
+  public static final int DEFAULT_CONCURRENCY_LEVEL = 0;
+  /**
```
alkemist: no need for javadoc, as this guy is private

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+   * If defined, run only health check with the given name.
+   */
+  private final String name;
+
+  /**
```
alkemist: same here.

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+ *   }
+ * }
+ * }</pre>
+ *
+ * @see ratpack.health.HealthCheck
+ * @see ratpack.health.HealthCheckResults
+ * @see ratpack.health.HealthCheckResultsRenderer
+ */
+public class HealthCheckHandler implements Handler {
+  public static final String DEFAULT_NAME_TOKEN = "DEFAULT";
```
alkemist: "name" might be a better value for this.

Should also have Javadoc.

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+  /**
+   * Run individual health check with the given name or all health checks.
+   * The {@code concurrencyLevel} determines parallelism in health check execution.
+   *
+   * @param context request context and exec control as well
+   * @throws Exception if anything goes wrong, exception will be implicitly passed to the context's {@link Context#error(Throwable)} method
+   */
+  @Override
+  public void handle(Context context) throws Exception {
+    if (!name.equals(DEFAULT_NAME_TOKEN)) {
```
alkemist: This logic doesn't seem right.

What we want to support is running a particular check if its name was specified as a path token. This classes param should indicate the name of the path token to look for. If the token is present, we only render the check whose name matches.

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+      handleByName(context, name);
+    } else {
+      handleAll(context);
+    }
+  }
+
+  /**
```
alkemist: don't need the javadoc for private methods.

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+        hcheckResults.put(hcheck.get().getName(), r);
+        context.render(new HealthCheckResults(ImmutableSortedMap.<String, HealthCheck.Result>copyOfSorted(hcheckResults)));
+      });
+    } catch (Exception ex) {
+      hcheckResults.put(hcheck.get().getName(), HealthCheck.Result.unhealthy(ex));
+      context.render(new HealthCheckResults(ImmutableSortedMap.<String, HealthCheck.Result>copyOfSorted(hcheckResults)));
+    }
+  }
+
+  /**
```
alkemist: don't need the javadoc for private methods.

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+    Optional<HealthCheck> hcheck = context.first(TypeToken.of(HealthCheck.class), hc -> hc.getName().equals(name));
+    if (!hcheck.isPresent()) {
+      context.clientError(404);
+      return;
+    }
+    try {
+      Promise<HealthCheck.Result> promise = hcheck.get().check(context.getExecution().getControl());
+      promise.onError(throwable -> {
+        hcheckResults.put(hcheck.get().getName(), HealthCheck.Result.unhealthy(throwable));
+        context.render(new HealthCheckResults(ImmutableSortedMap.<String, HealthCheck.Result>copyOfSorted(hcheckResults)));
```
alkemist: Can use Collections.singletonMap here instead of a concurrent map and then copying.

Same with the similar lines.

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
```diff
@@ -0,0 +1,435 @@
...
+            });
+          }
+        }
+      });
+    }).then(results -> {
+      context.render(new HealthCheckResults(ImmutableSortedMap.<String, HealthCheck.Result>copyOfSorted(hcheckResults)));
+    });
+  }
+
+  /**
```
alkemist: don't need javadoc

ratpack-core/src/main/java/ratpack/health/HealthCheckHandler.java
alkemist: Actually, I guess we need the ImmutableSortedMap.of()

PR #616, opened 2015-03-31
CsvReporter should only be provided when configuration is valid

### LINKED ISSUE #610: Exception when launching from jar and using CodaHaleMetricsModule
I am trying to start a server when `Ratpack.groovy` is inside a jar. Using the CodaHaleMetricsModule causes an `UnsupportedOperationException` to get thrown.

I forked the example gradle app and added shadow and the metrics to it. You can reproduce the stacktrace using `./gradlew runShadow`: https://github.com/danthegoodman/example-ratpack-gradle-groovy-app/tree/metrics_and_shadow

Ratpack Version: `0.9.14`
JVM: `1.8.0_31 (Oracle Corporation 25.31-b07)`
OS: `Mac OS X 10.10.2 x86_64`

According to the stacktrace, the error comes from the `CsvReporterProvider`:

```text
1) Error injecting constructor, java.lang.UnsupportedOperationException
  at ratpack.codahale.metrics.internal.CsvReporterProvider.<init>(CsvReporterProvider.java:37)
  while locating ratpack.codahale.metrics.internal.CsvReporterProvider
  at ratpack.codahale.metrics.CodaHaleMetricsModule.configure(CodaHaleMetricsModule.java:465) (via modules: com.google.inject.util.Modules$OverrideModule -> com.google.inject.util.Modules$OverrideModule -> com.google.inject.util.Modules$OverrideModule -> ratpack.codahale.metrics.CodaHaleMetricsModule)
  while locating com.codahale.metrics.CsvReporter

...

Caused by: java.lang.UnsupportedOperationException
  at com.sun.nio.zipfs.ZipPath.toFile(ZipPath.java:597)
  at ratpack.codahale.metrics.internal.CsvReporterProvider.<init>(CsvReporterProvider.java:40)
```

danthegoodman:
Fixes #610 

In the issue, I mentioned throwing an exception if the CSV config was not present. Instead, I am returning null if not configured. This is because the CsvReporter binding is marked as a Singleton, and guice singletons are eager singletons. Throwing an exception like I mentioned makes a csv configuration required.

That said, binding to null means that all attempts to get the CsvReporter instance when the csv configuration is not enabled will return null instead of providing a clear error message.

The alternative to returning null is to make the CsvReporterProvider a singleton and have it lazily create the CsvReporter singleton. If you would prefer that approach, I can make the necessary changes

### DIFF
--- a/ratpack-codahale-metrics/src/main/java/ratpack/codahale/metrics/internal/CsvReporterProvider.java
+++ b/ratpack-codahale-metrics/src/main/java/ratpack/codahale/metrics/internal/CsvReporterProvider.java
@@ -19,11 +19,9 @@ package ratpack.codahale.metrics.internal;
 import com.codahale.metrics.CsvReporter;
 import com.codahale.metrics.MetricRegistry;
 import ratpack.codahale.metrics.CodaHaleMetricsModule;
-import ratpack.server.ServerConfig;
 
 import javax.inject.Inject;
 import javax.inject.Provider;
-import java.io.File;
 
 /**
  * A Provider implementation that sets up a {@link CsvReporter} for a {@link MetricRegistry}.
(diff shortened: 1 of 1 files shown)


rhart:
Thanks for the PR @danthegoodman 

I think if we are unable to provide a sensible default then we need to make it so you can't construct a csv config without a reporting directory.  Probably with a constructor.  But we also need to make sure this works when using external config.

Also, I'm not sure that returning null is the way to go when no config is set.  The more I think about it the more I think reporters shouldnt be singletons at all.  I can imagine wanting to create multiple reporters of the same type with different regex and different reporting intervals for example.  Whilst you can't do this with our current approach I wonder whether a first step is to remove the singleton scope and in the onstart service (where we start the reporter) add it to the registry.  Or just remove the providers entirely and create and start the reporters based on the config in the onstart service.

Thoughts?

danthegoodman:
> I can imagine wanting to create multiple reporters of the same type with different regex and different reporting intervals for example.

If someone needs a tailored reporter or set of reporters, they may create their own Service and start whatever reporters they desire. I already use this approach for starting a Graphite Reporter. Maybe we could remove the csv configuration entirely and should just ask people to start a CsvReporter on their own, configured and programmed to their needs.

In either case, the only value I see to keeping the Reporters as Singletons is to provide access to the reporters outside of the Module. I think this sounds like exposing internal implementation and seems messy. If you keep the configurations for console, jmx, csv or whatnot, I am of the opinion that you should remove the bindings and providers and just start the reports in the onstart service.

PR #622, opened 2015-04-03
Add the maven reference to fix build.

chriseyre2000:
This will fix the documentation for building ratpack.

### DIFF
--- a/ratpack-manual/src/content/chapters/02-quick-start.md
+++ b/ratpack-manual/src/content/chapters/02-quick-start.md
@@ -89,6 +89,7 @@ apply plugin: "idea"
 
 repositories {
   jcenter()
+  maven { url 'http://clinker.netty.io/nexus/content/repositories/snapshots' }
 }
 
 dependencies {


rhart:
Thanks for this @chriseyre2000 but there should be no need to add the Netty snapshot repo anymore.  As of 0.9.15 we're on 4.1.0.Beta4.  Does the snippet not work for you on the latest version?

chriseyre2000:
No it does not work.
I am using a windows 7 box to build this and do need to make the suggested change to get the code to build on a clean system.
If I comment out the change it now works, but only because I now have the reference cached.

rhart:
And just double checking, this is with 0.9.15?

chriseyre2000:
Yes. I was using the groovy sample from the documentation:

buildscript {
  repositories {
    jcenter()
  }
  dependencies {
    classpath "io.ratpack:ratpack-gradle:0.9.15"
  }
}

apply plugin: "io.ratpack.ratpack-groovy"
apply plugin: "idea"

repositories {
  jcenter()
  maven { url 'http://clinker.netty.io/nexus/content/repositories/snapshots' }
}

dependencies {
  runtime "org.slf4j:slf4j-simple:1.7.10"
}

This has failed on two distinct windows boxes (one windows 7, the other windows 8.1) both only recently had java (&c) installed.

rhart:
Strange :s  I'm on Windows 7 too and this works for me.  Plus all our example projects have this repo removed and their CI builds are running fine.

So what error do you get?  Is it complaining it can't find 4.1.0.Beta4 in an of the repos?  Is this from the command line or are you using an IDE?

chriseyre2000:
I am building on the command-line.
 It is complaining it can't find 4.1.0.Beta4 in an of the repos

But once I add the one line an build once I no longer need to add it - which is making recreating the problem difficult.

rhart:
The 4.1.0.Beta4 dependencies are defo in jcenter https://jcenter.bintray.com/io/netty/netty-codec-http/4.1.0.Beta4/

Are you behind a proxy??  Maybe that's causing problems.

I'm going to close this pull request if that's ok as we don't need this manual change.  If you're still having problems can you create a topic in the forum.  Thanks.

Issue #619, opened 2015-04-02
Add support for fragments in the Thymeleaf module


tomakehurst:
I'd like to be able to return an HTML fragment from an XHR request.

Thymeleaf supports supplying an `IFragmentSpec` parameter when processing templates so it would seem to make sense to add this as an optional property on `Template`, possibly with a constructor overload to pass a String fragment spec.

If you think this is reasonable I'm happy to submit a PR.

alkemist:
Change makes sense to me.  +1

beckje01:
Closing this as the PR was merged already.

PR #648, opened 2015-04-20
Add support for specifying Thymeleaf template fragments

tomakehurst:
As discussed in #619, this adds overloads to ``Template.thymeleafTemplate()`` allowing a fragment spec to be passed, supporting e.g. partial HTML returned to an XHR call.


alkemist:
Many thanks, as soon as we get our build fixed I'll pull this in.

alkemist:
Thanks!

Issue #650, opened 2015-04-20
HttpClient does not decompress responses


kewne:
When submitting a request with ```Accept-Encoding``` HTTP headers and the response does come compressed, calling ```ReceivedResponse.getBody.getText()``` yields garbled (compressed) text.

I've worked around this by subclassing DefaultHttpClient and ContentAggregatingRequestAction in my project to add Netty's HttpContentDecompressor to the pipeline, but I think having support for this directly in Ratpack would the best behavior, since the current one is a bit unexpected (in my case, I was proxying requests and missed it because I wasn't setting the header explicitly).

I can make a pull request with my fix (a one-liner), although I'd probably need to work out some tests too.

rhart:
A PR with some tests would be great!  

We also need to consider what we do with streaming as well.  And, if we are proxying and the client sends `Accept-Encoding` do we want to decompress just to compress again?

kewne:
I missed that decompress to compress scenario...
How about treating it as two separate scenarios?

1. Requests that are used by the app itself.
These should always perform decompression if the response comes with a ```Content-Encoding``` header (at least I can't think of a reason not to).
This is essentially what is implemented and my patch would handle it.
2. Requests that are piped directly to the response sent to the original client, which might not need decompression.

I'm not sure every application will need option 2 though that's my use case.
What I'm trying to do is proxy two systems so I can "copy"  the request and response, like an aspect or interceptor.
The way I would go at this would be to create special proxying methods on ```HttpClient``` (or something like a ```ProxyingEnabledHttpClient```), what do you think?

rhart:
There's another proxy use case.  Original client doesn't accept gzip, our app requests gzip (because we can deal with it) and has to decompress before sending back.

I wonder whether we can just add it as an option to `RequestSpec`.  By default we would decompress the response but we could do
```
      get { HttpClient httpClient ->
        httpClient.requestStream(otherAppUrl("foo2")) {
          it.decompress(false)
        } then { StreamedResponse responseStream ->
          responseStream.send(response)
        }
      }
```

What I'm not sure about is what happens when we try to send an already compressed response.  Whether we try to double compress.  Can't remember how that works exactly.

kewne:
That would fall under my second scenario.
To make it clearer, I'm thinking of something like this:

    get { HttpClient httpClient ->
        httpClient.requestAndPipe(response, otherAppUrl("foo2")) {
            // change request headers as usual
        } then { StreamedResponse response ->
[...]

PR #651, opened 2015-04-23
#650 Make HttpClient support compressed responses

kewne:


ratpack-core/src/test/groovy/ratpack/http/client/HttpProxySpec.groovy
```diff
@@ -193,4 +196,32 @@ transfer-encoding: chunked
...
+    given:
+    def stringBuf = new ByteArrayOutputStream()
+    def writer = new OutputStreamWriter(new GZIPOutputStream(stringBuf))
+    writer.write("bar")
+    writer.close()
+
+    otherApp {
+      get("foo") {
+        response.headers.add(HttpHeaderConstants.CONTENT_ENCODING, "gzip")
+        response.send(stringBuf.toByteArray())
```
rhart: I don't think we need to do this.  If we add the accept-encoding header to the request then the "otherApp" should deal with the zipping and adding the correct header.

rhart:
Thanks!  Do you fancy looking at the streaming too? :)

kewne:
Sure, not sure why I missed it...

kewne:
Hmmm, I also just realized this breaks ```receivedResponse.send(response)```, so it'll need some more work... Good news is streaming is working too :)

alkemist:
Looks like there are some failing tests in ratpack-core with this change.

alkemist:
@kewne is this “done” if so, would you mind looking at the failing tests?

kewne:
Sure, I just haven't had the time to go through this yet.
From what I've seen, the ```HttpContentDecompressor``` I've added also changes the ```Content-Encoding``` header. Since ```DefaultHttpClient``` is used in the tests too, the tests fails because the headers don't match...
I'll try to think of a way to get around this, let me know if you have any ideas.

kewne:
Finally got some time to finish this...

rhart:
Great!  Shout if you need anything

kewne:
I think it's finished.
In the end, I decided to add a ```RequestSpec.decompressResponse``` method to enable/disable on a per request basis, like you suggested.

Let me know if there's anything else I should add.

rhart:
Awesome.  I will take a look soon as poss.

ratpack-core/src/main/java/ratpack/http/client/internal/ContentAggregatingRequestAction.java
```diff
@@ -45,6 +46,9 @@ public ContentAggregatingRequestAction(Action<? super RequestSpec> requestConfig
 
   @Override
   protected void addResponseHandlers(ChannelPipeline p, Fulfiller<? super ReceivedResponse> fulfiller) {
+    if (requestSpecBacking.isDecompressResponse()) {
```
rhart: Can this go in `RequestActionSupport`

ratpack-core/src/main/java/ratpack/http/client/internal/ContentStreamingRequestAction.java
```diff
@@ -61,6 +62,9 @@ public ContentStreamingRequestAction(Action<? super RequestSpec> requestConfigur
 
   @Override
   protected void addResponseHandlers(ChannelPipeline p, Fulfiller<? super StreamedResponse> fulfiller) {
+    if (requestSpecBacking.isDecompressResponse()) {
```
rhart: Can this go in `RequestActionSupport`

ratpack-core/src/main/java/ratpack/http/client/internal/RequestActionSupport.java
```diff
@@ -54,7 +54,7 @@
   private final String host;
   private final int port;
   private final MutableHeaders headers;
-  private final RequestSpecBacking requestSpecBacking;
+  final RequestSpecBacking requestSpecBacking;
```
rhart: We shouldn't need to do this if adding the decompressor is done in this class

ratpack-core/src/main/java/ratpack/http/client/internal/RequestSpecBacking.java
```diff
@@ -51,6 +52,7 @@ public RequestSpecBacking(MutableHeaders headers, URI uri, ByteBufAllocator byte
     this.byteBufAllocator = byteBufAllocator;
     this.requestParams = requestParams;
     this.bodyByteBuf = byteBufAllocator.buffer(0, 0);
+    this.decompressResponse = false;
```
rhart: I think we should default to true, this will be the most common case

ratpack-core/src/test/groovy/ratpack/http/client/HttpClientSmokeSpec.groovy
```diff
@@ -517,4 +519,73 @@ BAR
...
+  def "can handle compressed responses"() {
+    given:
+    def stringBuf = new ByteArrayOutputStream()
+    def writer = new OutputStreamWriter(new GZIPOutputStream(stringBuf))
+    writer.write("bar")
+    writer.close()
+
+    otherApp {
+      get("foo") {
+        response.headers.add(CONTENT_ENCODING, "gzip")
```
rhart: We shouldn't need to do this or the GZipping.  If the client sends `Accept-Encoding` of `gzip` the response from this app should automatically get compressed

ratpack-core/src/test/groovy/ratpack/http/client/HttpClientSmokeSpec.groovy
```diff
@@ -517,4 +519,73 @@ BAR
...
+        response.send(stringBuf.toByteArray())
+      }
+    }
+
+    and:
+    handlers {
+      get { HttpClient httpClient ->
+        httpClient.request(otherAppUrl("foo")) { RequestSpec rs ->
+          rs.headers.set("Accept-Encoding", "compress, gzip")
```
rhart: If decompress is on, I wonder whether we should automatically set these headers if they're not already set

rhart:
Sorry for taking so long to review.  If you're happy to address the comments that'd be great, otherwise I'll take a look as part of merging.  Thanks again for this.

ratpack-core/src/main/java/ratpack/http/client/internal/ContentStreamingRequestAction.java
kewne: It can, great suggestion. I'll take care of it.

ratpack-core/src/main/java/ratpack/http/client/internal/RequestSpecBacking.java
kewne: That breaks backwards compatibility and a few tests, so I wanted to avoid it :stuck_out_tongue:
If it's not a big concern to you (understandably, it's not even 1.0 at this point) let me know.

ratpack-core/src/test/groovy/ratpack/http/client/HttpClientSmokeSpec.groovy
kewne: I couldn't find the code where this is being done, can you give me some pointers?

One thing I'm worried about is that, by not making explicit that the response is compressed, ```otherApp```'s response could be uncompressed, which would make the test pass incorrectly.

ratpack-core/src/main/java/ratpack/http/client/internal/RequestSpecBacking.java
rhart: I would say now is the time, before 1.0, to get this how we want.  Let's take advantage of not being concerned about backwards compatibility while we still can :)

alkemist: +1

ratpack-core/src/test/groovy/ratpack/http/client/HttpClientSmokeSpec.groovy
rhart: This is where it's being done https://github.com/ratpack/ratpack/blob/877b7a2b6e465944601712a92410525314c3a5da/ratpack-core/src/main/java/ratpack/server/internal/SmartHttpContentCompressor.java

I can't seem to find any tests specifically for this but there are some for file compression.

I do take your point though.  My concern is that I don't know what happens if the client sends the `Accept-Encoding` header in this case, will otherApp try to apply compression to an already compressed response??  I guess not or your tests wouldn't be passing.  Maybe we need to put other assertions in along the way to make sure the behaviour is as we expect.

kewne: Ok, I actually tried setting the header to do compression which, oddly enough, disables compression.
It worked when I just set ```Accept-Encoding``` on the client side and left the server side alone.

kewne:
Not sure if you guys receive notifications when I push so just making sure.
I think I've addressed all of your comments but let me know if there's some other change to be done.

rhart:
Great!  Can you add yourself to the contributors list (https://github.com/ratpack/ratpack/blob/master/ratpack-manual/src/content/chapters/99-about-the-project.md) and I'll take a look over the weekend

rhart:
Thanks again for this @kewne, great stuff!

Issue #504, opened 2014-11-25
Add a cookie based session store


danhyun:
`ratpack-session` currently supports a memory based session store, we should provide a `CookieBasedSessionStore` and `CookieBasedSessionStorage` as an alternative to MapBasedSession implementations.

danhyun:
2 has been addressed. Still need to discuss 4.

wrt4: Doesn't make sense/is not practical to serialize object graphs, is `toString()` too simplistic? I'm not aware of any middle ground. Does it make sense to throw `IllegalArgumentException` if you don't explicitly pass a string as both key and value?

alkemist:
`toString()` is not round-trippable. That is, you can't recreate the object from the string. 

It's certainly an option to only support strings when using client side storage. We can at least start there. Have to make sure this limitation is prominently documented.

danhyun:
We could offer to serialize/deserialize to JSON, I would rather not introduce Jackson as a dependency to sessions but it would be nice to treat the client session store as a true multi level map.

alkemist:
We wouldn't implicitly convert to JSON. The user would either need to tell us to do this (by way of a bi directional converter) or do it before giving us the value (i.e. store the JSON representation directly)

alkemist:
We should have a look at the new Spring Session API for inspiration.

ylemoigne:
Note : the serialization problem made ClientSideSessionsModule not compatible with pac4j at this time.
(as the store return the "toString" of registered UserProfile, it end with ClassCastException in Pac4jProfileHandler.java:50)

danhyun:
Thanks to @zedar for addressing a bulk of the remaining issues as well as implementing really nice features. I'm going to close this ticket to mark that we have a solid basic client side session store but I'm going to open a new ticket to address serialization and pac4j here #664.

alkemist:
Docs need review.

PR #656, opened 2015-04-26
Cookie session support for Path, Domain and max cookie size (partitions).

zedar:
#504 - ```ClientCookieSessionModule.Config``` provides new attributes:

* ```path``` - session cookies will be attached to the given path. Default value is ```"/"```. 
* ```domain``` - session cookies will be attached to the given domain. Defalut values is ```null```
* ```maxCookieSize``` - if serialized and encoded session exeeds ```maxCookieSize``` it will be divided into partitions. Default value is ```2048```. Custom value has to be between ```1024``` and ```4096```.


danhyun:
Looks great, let's get this merged now and keep working at it.

PR #671, opened 2015-05-07
Update 70-heroku.md

saanvik:
Moving the gradle wrapper discussion and adding the command for creating the wrapper (the lazybones template doesn't include it, so it seemed like a good idea).

Also adding a link to foreman for pre-deployment testing.


rhart:
The lazybones template should defo include the gradle wrapper. Would you mind double checking?

saanvik:
You're right; sorry about that.

rhart:
No worries, thanks for taking the time anyway

Issue #664, opened 2015-05-01
Provide serialization options for ClienSideSessionModule


danhyun:
As mentioned in #504, pac4j does not work if the value stored in the session is a String so we should provide different serialization options to the user based on the user's requirements.

We should provide a Java serialization option as well as a String serialization option. As a part of this effort we should write tests to ensure that the Java serialization will enable good integration with pac4j.

danhyun:
We need to run some tests to determine how much bloat Java Serialization will add to the overall payload size.

alkemist:
Java serialization is very heavy and risky. I think we need to come with a custom serialization strategy out of the gate. We should not be allowing arbitrary objects to be serialized, so it's perfectly ok if we constrain this by requiring that a custom serialization/deserialization strategy be provided for each type you want to serialize.

davidmc24:
Maybe default to Jackson serialization?

alkemist:
Superseded by #681.

PR #674, opened 2015-05-13
Cookie session: serializing/deserializing the values of session entries with Java or String serializer   

zedar:
Defaults to `JavaValueSerializer` that writes/reads java objects to/from `ObjectOutputStream` and encodes/decodes bytes with `Base64` encoder/decoder. 

The `StringValueSerializer` converts java objects to and from `String`.

It is possible to define custom Serializer as the configuration parameter: `ClientSideSessionsModule.Config.valueSerializer`.

`Pac4j` sessions could now use `ClientSideSessionModule`. There are 2 parameters required: `valueSerializer` set to `JavaValueSerializer` and `secretToken` set to same fixed value. `secretToken` is required to support stateless clustering of ratpack servers.


zedar:
#664

alkemist:
@zedar I think you should rebase this and then merge. We can do follow on work separately.

PR #676, opened 2015-05-14
Javadoc method name fix

bonifaido:


alkemist:
Hi @bonifaido 

It's actually correct the way it is due to a change in 0.9.17. 

All of those code samples are actually compiled and executed as part of the build.

Thanks for taking the time to send it in.

Issue #669, opened 2015-05-07
OSGi support


almibe:
Hey all.  Has the topic of adding OSGi support to ratpack come up before?  I didn't check all of ratpack's dependencies, but the ones that I did check (over half and including netty) all had OSGi metadata in their manifests.  If there is interest in supporting OSGi I'd be willing to spend some time working on the issue.

Edit: By adding OSGi support I just mean adding OSGi metadata to the manifests at build time.  There are (at least) two common ways to do this with gradle.

alkemist:
It hasn't come up. We should support it unless it's very onerous. I don't really know what's involved.

almibe:
I should be able to test out adding OSGi metadata at build time this week.  It's usually trivial for libraries since you normally just want to export everything and don't want to take an active role in the lifecycle of the OSGi bundle.  The best case scenario is a one liner (https://github.com/TomasMikula/EasyBind/commit/409202070c279956f22ea83c8fb46db87575231f).  I'm not sure if the groovy extension classes in core will still work or not but other than that I don't see any issues.

almibe:
I was testing this out and I ran into an issue since one of ratpack's dependencies isn't an OSGi bundle, netty-transport-native-epoll-4.1.0.Beta4-linux-x86_64.  Which is weird since all the other netty jars have the OSGi metadata in them.  Since when it comes to library dependencies OSGi is all or nothing without a bunch of hacks, for the time being adding OSGi support to ratpack isn't possible without ugly workarounds.  I'm going to put this on the back burner (and for now just put the workaround hacks in my build script) so feel free to either leave this issue open or close it as you see fit.

almibe:
Actually after thinking this over some more I think I'm going to go ahead and make a pull request with OSGi support added to the gradle builds for the core and pac4j subprojects.  I think even with the issue mentioned above it is worthwhile to have the OSGi metadata in place since it will make the workarounds I mentioned simpler and when the netty issue is resolved we'll be good to go.

I couldn't find a good way to add OSGi support to all of the subprojects so I think doing it on a case by case basis is the way to go.  I'm currently only using ratpack-core in my application, but I plan on using the pac4j project eventually and it looks like all the pac4j stuff is already packaged as OSGi bundles on their side.

stevengssns:
Hi,

has anyone had another look at providing OSGi support?

[...]

PR #677, opened 2015-05-15
OSGi support for core and pac4j subprojects

almibe:
#669


alkemist:
Thanks @almibe ! 

Will take a look over the weekend.

alkemist:
I don't really like the idea of just partially doing this. Seems a bit arbitrary.

Thanks for taking the time to investigate and send the patch. I'm going to close this off until that Netty issue you raised gets addressed.

almibe:
That's fair.  I think they are wanting that fixed in the next version.  So when that happens I'll make another pull request.

Issue #681, opened 2015-05-25
Session support redesign


alkemist:
We need to rethink the way we are handling sessions.

We can simplify this to:

> A session is a blob that can be stored somewhere, and read as Java objects during request processing. During request processing, the session data can also be updated.

Some assumptions:

1. Session usage is read oriented
2. Session data is not large

Goals:

1. Make it easy to drop in new adapters that store/load the session data as a binary blob (i.e. they do not need to interpret the data at all)
2. Make the strategy for converting the session data to/from Java objects user defined (with some default impls)
3. Avoid redundant writes (i.e. don't write the session data if it hasn't changed)
4. Allow types to be used as keys, like `Registry`
5. Take care of session id generation

Non-goals:

1. Support patching the session data in the store (i.e. if anything changes, the whole session is written out)
2. Detecting changes to hydrated objects (i.e. changes must be explicit, by writing back the changed object)
3. Detect outdated writes (i.e. last write wins)

Unclear:

1. Session lifecycling (e.g. idle timeouts?)

### Implementation

```language-java
// to be implemented by stores (e.g. client side, local memory, redis)
interface SessionStoreAdapter {
  Promise<Boolean> store(String sessionId, ByteBuf sessionData)
  Promise<ByteBuf> load(String sessionId)
  Promise<Boolean> remove(String sessionId)
}
```

```language-java
// Strategy for turning objects into bytes and vice versa
// We provide default Java serialization and Jackson impls
interface SessionValueSerializer {
  <T> ByteBuf serialize(ByteBufAllocator bufferAllocator, Class<T> type, T value);
  <T> T deserialize(ByteBufAllocator bufferAllocator, Class<T> type, ByteBuf data);
}
```

```language-java
// User interface to the session
// Loaded lazily 
interface Session {
  Optional<String> get(String key);
  <T> Optional<? extends T> get(Class<T> key);
  <T> Optional<? extends T> get(Class<T> key, SessionValueSerializer serializer);

  Optional<String> set(String key, String value);
  <T> Optional<? extends T> set(Class<T> key, T value);
  <T> Optional<? extends T> set(Class<T> key, T value, SessionValueSerializer serializer);
  <T> Optional<? extends T> set(T value);
  <T> Optional<? extends T> set(T value, SessionValueSerializer serializer);

  Set<String> getStringKeys();
  Set<Class<T>> getTypeKeys();

  Optional<String> remove(String key);
  <T> Optional<? extends T> remove(Class<T> key);

[...]

PR #689, opened 2015-06-01
Client side session redesign (cookie based) - add inactivity interval verification

zedar:
#681


ratpack-session/src/main/java/ratpack/session/clientside/internal/ClientSideSessionStore.java
```diff
@@ -97,6 +105,54 @@ public Operation remove(AsciiString sessionId) {
...
+  }
+
+  private void setLastAccessTime() {
+    ByteBuf data = null;
+    try {
+      data = Unpooled.buffer();
+      data.writeLong(System.currentTimeMillis());
+      int oldCookiesCount = getCookies(config.getLastAccessTimeCookieName()).length;
+      String[] partitions = serialize(data);
+      for (int i = 0; i < partitions.length; i++) {
```
alkemist: why would there be more than one access time cookie?

alkemist: Ah of course, we are encrypting and signing which may blow out the size.

Issue #690, opened 2015-06-02
Support httpOnly and secure cookie for ratpack-session


asv:
Hi,

I can't find a way to set the HttpOnly config flag for the creation of a secure session cookie for ClientSideSession / CookieBasedSessionId. Netty Cookie has support this capabilities and would be cool expose it.

Thanks for Ratpack!

zedar:
@asv : good point! We definitely should add `HttpOnly` and `Secure` params to `ClientSideSessionConfig` as well as to `SessionIdCookieConfig`.

asv:
Also about session... Why is session generated (and stored) on server (*JSESSIONID*) when ClientSideSession is active (and *ratpack_session_ID* is sent)?

alkemist:
Not sure what you mean by stored here.

We still generate the session id so that it's available to log if the app wishes, for correlation.

zedar:
Changes addressed in 0.9.18-SNAPSHOT. Can we close this issue?

alkemist:
Yes, please close.

PR #691, opened 2015-06-02
CookieSession: expose HttpOnly and Secure attributes.

zedar:
#690 - `ClientSideSessionConfig` and `SessionIdCookieConfig` exposes `HttpOnly` and `Secure` attributes.
Defaults to: HttpOnly = true; Secure = false


alkemist:
I wonder if we should try and somehow share these settings between the id and data cookies. Might be more complication than it's actually worth.

LGTM. +1 for merge.

zedar:
I wonder if attributes like path, domain could be different between JSESSIONID and ratpack session cookies? Probably there is no reason. If not we can make `ClientSideSessionConfig` extension of `SessionIdCookieConfig`. What is your opinion?

alkemist:
I think I'm leaning towards making them the same. In this case, we'd rename `SessionIdCookieConfig` to `SessionCookieConfig` and leave `ClientSideSessionConfig` with the same name. The client side session module would just depend on `SessionCookieConfig` for generic cookie related settings (e.g. httpsOnly), but keep the crypto settings for example.

zedar:
Then `ClientSideSessionStore` will have a reference to 2 configs, `SessionCookieConfig` for basic and common attributes and `ClientSideSessionConfig` for more specific attributes. I think I can make such change today/tomorrow.

alkemist:
ClientSideSessionStore will simply depend on `SessionCookieConfig`, like it essentially does. That is, the provider method for `ClientSideSessionStore` which just take it as a param.

zedar:
ClientSideSessionStore depends on ClientSideSessionConfig and needs an access to common attributes (path, domain, httponly, secure) as well as client side specific (maxCookieSize, secureToken, ...). If we have 2 configs, one set by SessionModule and the other set be ClientSideSessionModule, then ClientSideSessionStore should have reference to both configs.

zedar:
@alkemist could you please have a look at the last commit?

ratpack-session/src/main/java/ratpack/session/SessionCookieConfig.java
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  /**
+   * The name of the cookie for session id.
+   * <p>
+   * Defaults to: {@code JSESSIONID}
+   *
+   * @return the name of the cookie for session id
+   */
+  public String getName() {
```
alkemist: Need to rename this to make it clear it's for the session id cookie name.

`getIdName()` ?

alkemist:
LGTM, besides SessionCookieConfig#getName()

zedar:
`getIdName()` is much better name. Will change it and merge to master.

PR #696, opened 2015-06-09
Add Spring Boot native support via @EnableRatpack

dsyer:
User can now (optionally) embed Ratpack as a server in a Spring
Boot application natively (as opposed to the other way round
which was already provided by Spring.spring()). Adding
@EnableRatpack should be enough to get it working with static
resources in classpath:static/ or classpath:/public (just like
in other Spring Boot servers). To use existing Ratpack guice modules
just add beans of type Module, e.g. for JSON add a @Bean of type
JacksonModule, etc.

Most of the features of dsyer/spring-boot-ratpack are here, except the
autoconfiguration (Jackson and Groovy templates) and the Groovy DSL
for Spring Boot CLI.

### DIFF
--- a/ratpack-spring-boot/src/main/java/ratpack/spring/config/internal/ChainConfigurers.java
+++ b/ratpack-spring-boot/src/main/java/ratpack/spring/config/internal/ChainConfigurers.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2012-2013 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ratpack.spring.config.internal;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.annotation.AnnotationAwareOrderComparator;
+import org.springframework.stereotype.Component;
+
+import com.google.common.collect.ImmutableMap;
+
+import ratpack.func.Action;
+import ratpack.handling.Chain;
+import ratpack.handling.Handler;
+import ratpack.handling.Handlers;
+import ratpack.path.PathBinder;
+import ratpack.path.PathBinding;
+import ratpack.path.internal.DefaultPathBinding;
+import ratpack.server.ServerConfig;
+import ratpack.spring.config.RatpackServerCustomizer;
+
+/**
+ * @author Dave Syer
+ *
+ */
+@Component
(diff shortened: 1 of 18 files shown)


dsyer:
I don't know but the CI failure looks like a network glitch to me. I guess if I just push again it will try again?

dsyer:
Also, the `ratpack-guice` dependency should be optional, but I don't know how to do that in gradle.

danhyun:
Hi @dsyer , thanks so much for this, I'll retrigger with a longer timeout

alkemist:
Thanks @dsyer 

Do you mind if we remove the `@author` tags from the code? We don't use them across the board on this codebase.

dsyer:
No, of course not. There *is* a copyright header that mentions "the original author". I thought they were correlated, but I'm happy to go with the flow.

ratpack-spring-boot/src/main/java/ratpack/spring/config/internal/ChainConfigurers.java
```diff
@@ -0,0 +1,98 @@
...
+    for (Action<Chain> delegate : delegates) {
+      if (!(delegate instanceof ChainConfigurers)) {
+        delegate.execute(chain);
+      }
+    }
+  }
+
+  private Action<Chain> staticResourcesAction(final ServerConfig config) {
+    return chain -> {
+      chain.all(Handlers.path(new RootBinder(), Handlers.assets(config, "static", Arrays.asList("index.html"))));
```
danhyun: Hi Dave,

What's the expected behavior here? To respond to any root context requests against the static and public directories?

ratpack-spring-boot/src/test/groovy/ratpack/spring/DefaultStaticResourceTests.java
```diff
@@ -0,0 +1,65 @@
...
+public class DefaultStaticResourceTests {
+
+  private TestRestTemplate restTemplate = new TestRestTemplate();
+
+  @Autowired
+  private RatpackServer server;
+
+  @Test
+  public void contextLoads() {
+    ResponseEntity<String> result = restTemplate.getForEntity("http://localhost:" + server.getBindPort() + "/main.css",
```
danhyun: Given the `staticResourcesAction` defined in `ChainConfigurers` this should first look in "static", then in "public", is this the expected behavior?

danhyun:
I've brought in changes, we can work on it from here.

dsyer:
When you say "brought in changes" you mean you modified the content but it's in master (see commit ff820a5). Just in case anyone wonders.

danhyun:
Yes sorry, I'm in transit I'll continue with this in a bit.

ratpack-spring-boot/src/main/java/ratpack/spring/config/internal/ChainConfigurers.java
dsyer: To respond to any unhandled requests against those directories, yes. It seems to work, but I'm not really 100% confident I know what ratpack's rules are for matching requests.

ratpack-spring-boot/src/test/groovy/ratpack/spring/DefaultStaticResourceTests.java
dsyer: Yes. It mimics what Spring Boot does with Servlet containers, so users migrating from Tomcat or Jetty (or reading the Boot user guide) will find this a comfortable default. If you prefer I can move it out to my sping-boot-ratpack project (which has some additional features).

danhyun:
Thanks for the contribution @dsyer !

PR #699, opened 2015-06-10
SSL: every SSLHandler instance has to have own SSLEngine instance.

zedar:


ratpack-core/src/main/java/ratpack/server/internal/DefaultRatpackServer.java
```diff
@@ -189,10 +189,7 @@ private ChannelHandler buildHandler(DefinitionBuild definitionBuild) throws Exce
   protected Channel buildChannel(final ServerConfig serverConfig, final ChannelHandler handlerAdapter) throws InterruptedException {
 
     SSLContext sslContext = serverConfig.getSSLContext();
-    if (sslContext != null) {
-      this.sslEngine = sslContext.createSSLEngine();
-      sslEngine.setUseClientMode(false);
-    }
+    this.ssl = sslContext != null ? true : false;
```
danhyun: whynot just `this.ssl = sslContext != null`

ratpack-core/src/main/java/ratpack/server/internal/DefaultRatpackServer.java
```diff
@@ -319,7 +318,7 @@ public synchronized boolean isRunning() {
 
   @Override
   public synchronized String getScheme() {
-    return isRunning() ? sslEngine == null ? "http" : "https" : null;
+    return isRunning() ? !this.ssl ? "http" : "https" : null;
```
danhyun: in this case I think
`return isRunning() ? this.ssl ? "https" : "http" : null;` reads cleaner

danhyun:
changes look good just waiting on snap

PR #700, opened 2015-06-11
Add some docs for the Spring support

dsyer:
The tests fail, but I have no idea how to work the snippet
fixtures, so I don't know if I can do anything about it. The
snippets all run and are valid apps though.

### DIFF
--- a/ratpack-manual/src/content/chapters/36-spring.md
+++ b/ratpack-manual/src/content/chapters/36-spring.md
@@ -1,0 +1,148 @@
+# Spring Boot
+
+The `ratpack-spring-boot` extension provides integration with [Spring Boot](http://projects.spring.io/spring-boot).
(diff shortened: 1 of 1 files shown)


danhyun:
Great, thanks so much. Our two ways of testing documented code are made available via `ManualCodeSnippetTests` and `JavadocCodeSnippetTests`. This would be picked up by `ManualCodeSnippetTests`. I can take a look later today to see what's happening.

danhyun:
Ok so it seems we may have a bug in the `Spring` integration code, particularly with https://github.com/ratpack/ratpack/blob/master/ratpack-spring-boot/src/main/java/ratpack/spring/Spring.java#L90

Seems we should be disable web mode so we can let Ratpack handle the requests instead of Spring Boot's embedded servlet container.

danhyun:
Thanks again Dave.

PR #704, opened 2015-06-15
Add pac4j documentation

leleuj:


danhyun:
@leleuj merci b'coup!

Just a few minor details I will comment in the diffs.

ratpack-manual/src/content/chapters/37-pac4j.md
```diff
@@ -0,0 +1,94 @@
+# Security using pac4j
+
+The `ratpack-pac4j` extension provides authentication support based on [pac4j](https://github.com/pac4j/pac4j).
+
+`pac4j` is a generic authentication engine which is implemented by many frameworks and supports multiple protocols: OAuth, CAS, OpenID (Connect), SAML, Google App Engine and HTTP (form and basic auth).
+
+The `RatpackPac4j` class is the main component to use to manage the security of your web application.
```
danhyun: It would be nice to link to the Javadoc url here i.e.

```
[`RatpackPac4j`](api/ratpack/pac4j/RatpackPac4j.html)
```

ratpack-manual/src/content/chapters/37-pac4j.md
```diff
@@ -0,0 +1,94 @@
...
+
+
+## Prerequisite
+
+All authentication mechanisms work by redirecting the user to an identity provider and finishing the login process when he is redirected back to the application with some dedicated credentials.
+To make these authentication processes work and to save the originally requested url, the session is required. That's why the `SessionModule` is necessary:
```
danhyun: Again would be nice to link to the Javadoc whenever possible.

ratpack-manual/src/content/chapters/37-pac4j.md
```diff
@@ -0,0 +1,94 @@
...
+
+```language-java
+.registry(Guice.registry(b -> b
```
danhyun: These `language-java` code snippets are tested and require a full Java code listing.

e.g.

import ratpack.guice.Guice;
import ratpack.session.Session;
import ratpack.session.SessionModule;
import ratpack.test.embed.EmbeddedApp;

import static org.junit.Assert.assertEquals;

public class Example {
  public static void main(String[] args) throws Exception {
    EmbeddedApp.of(s -> s
      .registry(
        Guice.registry(b -> b.module(SessionModule.class)))
      .handlers(c ->
        c.all(ctx -> {
          Session session = ctx.get(Session.class);
          ctx.getResponse()
            .send("Context contains session: " + (session != null));
        })))
      .test(testClient ->
        assertEquals("Context contains session: true", testClient.getText())
      );
  }
}

ratpack-manual/src/content/chapters/37-pac4j.md
leleuj: Yes. Will do.

ratpack-manual/src/content/chapters/37-pac4j.md
leleuj: I agree, I will add a Javadoc link for the `SessionModule` but one must be careful when refactoring happens and package names change...

danhyun: Absolutely, that's why we actually resolve links as a part of our build process. As a part of the `check` task, we scan all links in the manual and make sure they resolve. Our docs are fully tested and checked for broken links :)

leleuj: Perfect. No worries then!

ratpack-manual/src/content/chapters/37-pac4j.md
leleuj: OK. If so, I'd prefer to remove this `language-java` tag and just keep the code straightforward: the few lines of source code that are really relevant.

But when I remove the `language-java` tag, the source code is not highlighted anymore: is there a way to keep the highlight with a partial source code?

leleuj: I found a trick using the `language-bash` tag: is it acceptable for you?

danhyun: I'm not sure about this, but the case for using a whole code listing is to not only demonstrate the code usage but to also put it in a context. Our APIs are also pretty volatile at the moment so it helps to make sure that not only is the mechanism working as advertised, but working against the current state of the code.

Meaningful assertions also help to show that our piece of code achieved the desired result.

leleuj: I guess it's a polite "no" ;-)

I will use real examples and try to stay as close as possible from the ones given in the Javadoc.

alkemist:
@leleuj we should aim not to just replicate the javadocs in the manual. 

The manual should focus on what the point of ratpack/pac4j is and what someone can use it for, not how to use it. That's a hard line to walk, but that's our mental model.

leleuj:
Sure. But the `pac4j` page does not replicate the javadocs, it provides a comprehensive presentation of the `ratpack-pac4j` module capabilities, doesn't it?

For the source code examples, I just wanted to focus on the very relevant lines, but if I need to choose working examples, I think it's better to stay close to the ones provided in the Javadoc, isn't it?

ratpack-manual/src/content/chapters/37-pac4j.md
```diff
@@ -0,0 +1,94 @@
+# Security using pac4j
+
+The `ratpack-pac4j` extension provides authentication support based on [pac4j](https://github.com/pac4j/pac4j).
+
+`pac4j` is a generic authentication engine which is implemented by many frameworks and supports multiple protocols: OAuth, CAS, OpenID (Connect), SAML, Google App Engine and HTTP (form and basic auth).
```
alkemist: > which is implemented by many frameworks

Is this true? How would a framework _implement_ pac4j?

ratpack-manual/src/content/chapters/37-pac4j.md
```diff
@@ -0,0 +1,94 @@
...
+
+```language-java
+chain
+   .all(RatpackPac4j.authenticator(formClient, formClient, saml2Client, facebookClient, twitterClient, basicAuthClient, casClient))
+```
+
+The above command creates the `/authenticator` url endpoint to receive callbacks and finish the login processes (it can be set to an another `path` using `RatpackPac4j.authenticator(String path, Client<?, ?>... clients)`.
+
+
+## Url protection
```
alkemist: I'd be inclined to shorten this and the following sections, and just let the link to the javadoc do the job of providing the detail. Someone reading this page should be able to quickly understand that there are constructs for url protection, getting the profile etc. but it's perfectly ok for them to click through to the Javadoc to see detail.

ratpack-manual/src/content/chapters/37-pac4j.md
```diff
@@ -0,0 +1,94 @@
...
+All authentication mechanisms work by redirecting the user to an identity provider and finishing the login process when he is redirected back to the application with some dedicated credentials.
+To make these authentication processes work and to save the originally requested url, the session is required. That's why the `SessionModule` is necessary:
+
+```language-java
+.registry(Guice.registry(b -> b
+                  .module(SessionModule.class)
+```
+
+
+## Authentication mechanisms definition
```
alkemist: I think this section is one of the most important. Anybody using the pac4j stuff needs to understand well what `Client` is all about. The content here is pretty good, but I still think I'd be a little unclear about the abstraction after reading it.

ratpack-manual/src/content/chapters/37-pac4j.md
leleuj: I'm not sure to follow you. Should I provide some examples? `pac4j` is implemented in J2E (j2e-pac4j library), by the SSO CAS server (cas-server-support-pac4j module), in Play (play-pac4j library), in Vertx, Undertow, Sparkjava...

 `pac4j` is an authentication/authorization engine: it has many concepts and classes, a lot of common logic, a general algorithm which can be used by many frameworks after some adaptation like a dedicated `WebContext` for example.

Does it make more sense now?

ratpack-manual/src/content/chapters/37-pac4j.md
leleuj: I hoped it was clear enoguh: can you elaborate a little more about what's unclear? I don't want to enter into too many details, because for example, for the `FormClient` and the `BasicAuthClient`, I may have added that you need the additional `Authenticator` and `ProfileCreator` objects.

ratpack-manual/src/content/chapters/37-pac4j.md
leleuj: The Javadoc is great to know how you should exactly use a method or a class and from this point of view, I think that the Javadoc of the `RatpackPac4j` class is perfect.
But the Javadoc is not very good to give readers an overview and this is what this manual tries to do: what is the prerequisite, how you define the ways to login, how you protect an url, how you get the current user profile, how you logout.

My idea was to shorten the source code to make the user focus on the explanations and you want me to link more to the Javadocs.

Here is my proposal: remove the partial source codes to only keep the explanations and add more links to the Javadoc. Does it sound good to you?

leleuj: @alkemist : I'll be in holidays for 2 weeks. Just let me know how you want me to write the doc and I'll do that today if I have time or after my return...

leleuj:
@danhyun @alkemist I updated the doc to meet your recommendation: Javadoc links, working source code (not a copy/paste of the Javadoc), slight text rewrites. 

What do you think?

ratpack-manual/src/content/chapters/37-pac4j.md
```diff
@@ -0,0 +1,141 @@
...
+                ))
+                .handlers(chain -> {
+                    final FacebookClient facebookClient = new FacebookClient("fkey", "fsecret");
+                    final TwitterClient twitterClient = new TwitterClient("tkey", "tsecret");
+                    final FormClient formClient = new FormClient("/theForm.html", new SimpleTestUsernamePasswordAuthenticator(), new UsernameProfileCreator());
+                    final BasicAuthClient basicAuthClient = new BasicAuthClient(new SimpleTestUsernamePasswordAuthenticator(), new UsernameProfileCreator());
+                    final CasClient casClient = new CasClient();
+                    casClient.setCasLoginUrl("http://mycasserver/login");
+                    chain
+                        .all(RatpackPac4j.authenticator(formClient, formClient, facebookClient, twitterClient, basicAuthClient, casClient))
```
johnrengelman: Adding the `formClient` twice here isn't necessary

ratpack-manual/src/content/chapters/37-pac4j.md
```diff
@@ -0,0 +1,141 @@
...
+                        .module(TextTemplateModule.class)
+                        .module(SessionModule.class)
+                ))
+                .handlers(chain -> {
+                    final FacebookClient facebookClient = new FacebookClient("fkey", "fsecret");
+                    final TwitterClient twitterClient = new TwitterClient("tkey", "tsecret");
+                    final FormClient formClient = new FormClient("/theForm.html", new SimpleTestUsernamePasswordAuthenticator(), new UsernameProfileCreator());
+                    final BasicAuthClient basicAuthClient = new BasicAuthClient(new SimpleTestUsernamePasswordAuthenticator(), new UsernameProfileCreator());
+                    final CasClient casClient = new CasClient();
+                    casClient.setCasLoginUrl("http://mycasserver/login");
```
johnrengelman: I feel there's too much happening in this example. We don't need to show the usage of every client in this example. It should be kept to:
* How to configure a client and require authentication
* How to logout
* How to obtain and use the userProfile

ratpack-manual/src/content/chapters/37-pac4j.md
```diff
@@ -0,0 +1,141 @@
...
+import static java.util.Collections.singletonMap;
+import static ratpack.groovy.Groovy.groovyTemplate;
+
+public class Example {
+
+    public static void main(final String[] args) throws Exception {
+        RatpackServer.start(server -> server
+                .serverConfig(ServerConfig
+                        .baseDir(new File("src/main"))
+                        .port(8080)
```
johnrengelman: This `serverConfig` isn't necessary since it's the default.

ratpack-manual/src/content/chapters/37-pac4j.md
leleuj: Yes, good catch. Will remove.

ratpack-manual/src/content/chapters/37-pac4j.md
leleuj: OK. will clean.

alkemist:
I'm going to merge this in and do some work on it, then we can iterate from there.

leleuj:
Thanks. Just let me know if you need more help from me.

PR #705, opened 2015-06-16
Extended support for server and client SSL authentication

zedar:
1. Extend `SSLContexts` helpers with truststore settings. 
2. Add `SSLContext` to  `RequestSpec` used by ratpack's `HttpClient`. Now it is possible to define client side `SSLContext` used in ssl handshaking between server and client.  
3. Add `sslClientAuth` property to `ServerConfig`. If set to true, the server needs client ssl authentication. It propagates to the statement `sslEngine.setNeedClientAuth(sslClientAuth);` while building server's channel.


ratpack-core/src/main/java/ratpack/http/client/RequestSpec.java
```diff
@@ -40,6 +42,23 @@
...
+  RequestSpec sslContext(SSLContext sslContext);
+
+  /**
+   * Factory method to create {@link SSLContext} used for client and server SSL authentication.
+   *
+   * @param factory provides a factory that will create {@link SSLContext} instance
+   * @return the {@link RequestSpec}
+   * @throws Exception this can be thrown from the action
+   */
+  RequestSpec sslContext(Factory<SSLContext> factory) throws Exception;
```
alkemist: I'm not sure this is warranted, since we just invoke the factory immediately (i.e. a user can do this).

If you think it is needed, it should be implemented as an interface default method.

ratpack-core/src/test/groovy/ratpack/ssl/HttpsTruststoreSpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+      rs.sslContext {
+        return SSLContexts.sslContext(
+          HttpsTruststoreSpec.getResource("dummy.keystore"), "password",
+          HttpsTruststoreSpec.getResource("client_dummy.truststore"), "password")
+      }
+    }
+    get("foo")
+
+    then:
+    UncheckedIOException ex = thrown()
```
alkemist: Let's extract a method for this, it's duplicated.

ratpack-core/src/test/groovy/ratpack/ssl/HttpsTruststoreSpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    handlers {
+      get("foo") {
+        render "SSL VERIFIED"
+      }
+    }
+
+    when: "client does not provide valid certificate"
+    requestSpec { RequestSpec rs ->
```
alkemist: This method delegates the closure to RequestSpec. You don't need to specify the arg type, or the receiver of the method call.

ratpack-core/src/main/java/ratpack/server/ServerConfig.java
```diff
@@ -330,6 +337,14 @@ static Builder baseDir(File baseDir) {
     Builder ssl(SSLContext sslContext);
 
     /**
+     * The server needs client SSL authentication.
+     *
+     * @param sslClientAuth whether or not server needs client SSL authentication
+     * @return {@code this}
+     */
+    Builder sslClientAuth(boolean sslClientAuth);
```
alkemist: I think `requireClientSslAuth` is a better name for this.

ratpack-core/src/test/groovy/ratpack/ssl/HttpsTruststoreSpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+      }
+    }
+    get("foo")
+
+    then:
+    UncheckedIOException ex = thrown()
+    ex.getCause() instanceof SSLHandshakeException || ex.getCause() instanceof ClosedChannelException
+  }
+
+  def "throw exception when server does not provide truststore"() {
```
alkemist: If all request are going to fail with this config, we should fail sooner.

zedar: Exception is thrown during the handshake. Before client -> server call we do not know if keystore or trusture contains valid and required certificates or key pairs.

PR #706, opened 2015-06-18
Added Domain Specific Langauges to Groovy Topics

pieces029:
Would you want this to link to http://groovy-lang.org/dsls.html?


alkemist:
Thanks for the PR. I'm planning to rework this whole chapter soon, which will render this change obsolete. As such, I'm going to close it. 

Thanks for taking the time to submit.

Issue #517, opened 2014-12-09
Thymeleaf: Defect Using URL Expressions


sbonnick:
Previously discussed here: http://forum.ratpack.io/th-href-failures-with-Thymeleaf-td808.html

Using th:href, with a constant string set in the template: 

``` HTML
<link rel="stylesheet" type="text/css" href="css/bootstrap.min.css" th:href="@{thymeleaf/css/bootstrap.min.css}" />
```

Regardless of the string constant, the above fails with the following error: 

``` HTML
org.thymeleaf.exceptions.TemplateProcessingException: Link base "thymeleaf/css/bootstrap.min.css" cannot be context relative (/) or page relative unless you implement the org.thymeleaf.context.IWebContext interface (context is of class: org.thymeleaf.context.Context) (index:7)
```

alkemist:
> I believe I read somewhere that Ratpack does not depend on any javaEE libraries. I think this is a JavaEE library. Do we have a problem with bringing this in as a dependency? 

It's perfectly fine for the Thymeleaf module to depend on the Servlet API.

> Is there already a factory of sorts to create the request object?

No, we haven't needed to bridge to the servlet types before.

…

> So it seems we need to support:
> 1. getSession() -> and CRUD operations in the session
> 2. getParameterMap()
> 3. getContextPath()

I don't think we need to support writing to the Session. The view should not be doing this. The rest should be reasonably easy, but let's see.

> I'm thinking we may want to think about having a factory somewhere that any guice supported module can call to generate these objects based on the netty/ratpack request/context/response data.

This probably makes sense, but we can do it as a second step. It may make sense to extract the bridging to a `ratpack-servlet-api` type library. Let's see how successful we are with implementing them though first.

todd1251:
I tried the approach suggested by @sbonnick and it prevents the `TemplateProcessingException`. Almost everything can just throw `UnsupportedOperationException` with a few trivial implementations needed (context path as empty string i.e. root, request attributes as a map, encode URL as a no-op - but that might be wrong). Surely it is still broken in many ways...

A robust set of adapters to integrate Ratpack with any Servlet API based library would certainly be nice to have. That said, it does seem like a limitation in Thymeleaf to assume that the only possible web context is the servlet API, given it doesn't seem to use very much of it (alas exposing the request/response objects to the view).

[...]

PR #711, opened 2015-06-25
Workaround to support Thymeleaf link URL expressions (#517)

todd1251:


todd1251:
@Vadi: While it might be possible to exclude `servlet-api` when depending on Thymeleaf, it is a defined as a `provided` dependency in Thymeleaf's [pom.xml](https://github.com/thymeleaf/thymeleaf/blob/3.0-dev/pom.xml#L280) and used by the web context and link expression related classes. If `WebContext` is not used, then the exception noted in issue #517 is thrown. To exclude `servlet-api` completely, it would not be possible to use the link expressions (`@{...}`), unless a change is made upstream in Thymeleaf (ideally it would be *HTTP-engine-agnostic*). If someone really wanted to avoid shipping the `servlet-api` JAR then I suppose they still could, even with this change, by excluding the transitive dependency via Gradle and not using link expressions?

@alkemist: As I mentioned in the discussion of issue #517, this is probably still wrong/broken but it doesn't crash, at least. As *context path* is a servlet API concept, it being defined as an empty string here means it will never affect the resolved URL (it would otherwise be prepended). Not sure if some resolution is *supposed* to happen for relative URLs but it leaving them as-is seems like sensible behaviour to me, i.e. the browser will resolve the relative links. 

I added the test case based on the different types noted in the Thymeleaf docs on [Link URLs](http://www.thymeleaf.org/doc/tutorials/2.1/usingthymeleaf.html#link-urls) and was surprised to find that it's almost a no-op ;-).

alkemist:
Re servlet-api: Depending on the servlet-api jar is fine. Eventually, we'll have a `ratpack-servlet-adapter` thing that will depend on this and that thymeleaf depends on.

Re the links: so what exactly does this change practically give users?

todd1251:
@alkemist: The practical benefit is that the standard Thymeleaf syntax for writing links (`@{...}`) doesn't crash. It also doesn't do anything in most cases because the context path is always empty (given that is a servlet container thing).

It's okay if you don't want to accept the change and give it more thought first. I think, ideally, Thymeleaf wouldn't depend on the servlet API directly. It is possible to just leave out the `@{...}` and links will work fine but that is counter to the Thymeleaf documentation.

alkemist:
@todd1251 happy to merge this. Will do so later today.

Thanks.

johnrengelman:
Failure of this PR appeared to be a caching problem on Snap. I busted the cache and re-ran and it completed successfully.

Issue #176, opened 2013-10-11
Log reporting for Metrics


rhart:
Support reporting metrics to a log file

alkemist:
@rhart don't we already support this?

rhart:
We have support for console reporting, this is different.  Not critical for 1.0 as you can register your own log reporter, our integration just makes it easier to do so.

rhart:
Closed by https://github.com/ratpack/ratpack/pull/718

PR #718, opened 2015-07-03
SLF4J metrics reporter support

djjorjinho:
This request adds support for logging metrics to a Slf4j logger, optionally, instead of the console.

This comes in handy as a cheap way to do use Log collection and analysis software to broadcast notifications. 

Log collectors might be set up to read only log files produced by logging implementations and not STDOUT, thus the reason of this request.

Please let me know if there's anything else I can provide. Thank you.


rhart:
Great, thanks! This will address #176. I will take a look over the weekend. 

If you're looking for something else metric related then I guess graphite reporter is another common one we're missing. Also, a review and updates to the documentation around metrics would be very welcome. 

Failing that, let me know what kind of thing you are interested in and we'll find something else.

djjorjinho:
Thank you. I'm having a look at the Graphite reporter and thinking on how to tackle the various senders it supports. I'll send a request when I crack it.

ratpack-codahale-metrics/src/test/groovy/ratpack/codahale/metrics/MetricsSpec.groovy
```diff
@@ -73,6 +73,7 @@ class MetricsSpec extends RatpackGroovyDslSpec {
         .jmx()
         .csv { it.reportDirectory(reportDirectory.root).reporterInterval(Duration.ofSeconds(1)) }
         .console { it.reporterInterval(Duration.ofSeconds(1)) }
+        .slf4j { it.reporterInterval(Duration.ofSeconds(1)).prefix("test.") }
```
rhart: We need test assertions to make sure this has worked.  It should output to the `output` variable I think as Gradle configures simple logger for SLF4J which outputs to system out.  We should be able to identify this output via the prefix

ratpack-codahale-metrics/src/test/groovy/ratpack/codahale/metrics/MetricsSpec.groovy
```diff
@@ -476,6 +477,7 @@ class MetricsSpec extends RatpackGroovyDslSpec {
     bindings {
       module new CodaHaleMetricsModule(), {
         it.console { it.reporterInterval(Duration.ofSeconds(1)).includeFilter(".*ar.*").excludeFilter(".*bar.*") }
+        it.slf4j { it.reporterInterval(Duration.ofSeconds(1)).prefix("test.").includeFilter(".*ar.*").excludeFilter(".*bar.*") }
```
rhart: Same as the other test, we need to assert this is working as expected

rhart:
Sorry this has taken so long.  Can you also add yourself as a contributor in the manual section.  Thanks.

djjorjinho:
Not a problem. I've added the test assertions, but used a mock to print the output to sysout, using the logger() builder method in the process and making the test able to run outside of Gradle (IDE's and such.)

Let me know what you think.

rhart:
Great. I will get to this tomorrow

rhart:
Looks like there's a merge conflict in `ratpack-manual/src/content/chapters/99-about-the-project.md`.  Could you rebase and then this is good to go.

djjorjinho:
I did the rebase. Let me know if you need anything else.

PR #724, opened 2015-07-10
Graphite codahale metrics reporter Support

djjorjinho:
The contents of this request allows preliminary support for graphite metrics reporting, with a broad range of support for graphite senders.
I did not add the RabbitMQ Sender dependency under the assumption that the end user would add the dependency in their own projects if needed.

Let me know if there's anything that need changing and I'll get to it.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/724)

### DIFF
--- a/ratpack-codahale-metrics/src/main/java/ratpack/codahale/metrics/CodaHaleMetricsModule.java
+++ b/ratpack-codahale-metrics/src/main/java/ratpack/codahale/metrics/CodaHaleMetricsModule.java
@@ -20,20 +20,21 @@ import com.codahale.metrics.*;
 import com.codahale.metrics.Slf4jReporter.LoggingLevel;
 import com.codahale.metrics.annotation.Metered;
 import com.codahale.metrics.annotation.Timed;
+import com.codahale.metrics.graphite.GraphiteReporter;
 import com.codahale.metrics.jvm.GarbageCollectorMetricSet;
 import com.codahale.metrics.jvm.MemoryUsageGaugeSet;
 import com.codahale.metrics.jvm.ThreadStatesGaugeSet;
 import com.google.inject.Injector;
 import com.google.inject.matcher.Matchers;
 import com.google.inject.multibindings.Multibinder;
+import org.slf4j.Logger;
+import org.slf4j.Marker;
 import ratpack.codahale.metrics.internal.*;
 import ratpack.func.Action;
 import ratpack.guice.ConfigurableModule;
 import ratpack.handling.HandlerDecorator;
 import ratpack.server.Service;
 import ratpack.server.StartEvent;
-import org.slf4j.Logger;
-import org.slf4j.Marker;
 
 import javax.inject.Inject;
 import javax.inject.Provider;
@@ -159,6 +160,7 @@ public class CodaHaleMetricsModule extends ConfigurableModule<CodaHaleMetricsMod
     private Optional<WebSocket> webSocket = Optional.empty();
     private Optional<Csv> csv = Optional.empty();
     private Optional<Slf4j> slf4j = Optional.empty();
+    private Optional<Graphite> graphite = Optional.empty();
 
     /**
      * The state of jvm metrics collection.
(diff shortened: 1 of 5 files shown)


ratpack-codahale-metrics/src/main/java/ratpack/codahale/metrics/CodaHaleMetricsModule.java
```diff
@@ -651,6 +680,81 @@ public Csv reportDirectory(File reportDirectory) {
       }
     }
 
+    public static class Graphite {
+      private Duration reporterInterval = DEFAULT_INTERVAL;
+      private GraphiteSender sender;
+      private boolean enabled = true;
```
rhart: We also need the filters for this reporter.

ratpack-codahale-metrics/src/main/java/ratpack/codahale/metrics/CodaHaleMetricsModule.java
```diff
@@ -651,6 +680,81 @@ public Csv reportDirectory(File reportDirectory) {
       }
     }
 
+    public static class Graphite {
```
rhart: I'm wondering whether we should bring these out into their own class files, this module class is getting pretty big.  What do you think?

ratpack-codahale-metrics/src/test/groovy/ratpack/codahale/metrics/MetricsSpec.groovy
```diff
@@ -73,6 +74,7 @@ class MetricsSpec extends RatpackGroovyDslSpec {
         .jmx()
         .csv { it.reportDirectory(reportDirectory.root).reporterInterval(Duration.ofSeconds(1)) }
         .console { it.reporterInterval(Duration.ofSeconds(1)) }
+        .graphite { it.enable().sender(new Graphite("localhost", 2003)).reporterInterval(Duration.ofSeconds(1)) }
```
rhart: We need to test that this is actually working.  I think we should be able to use a test ratpack app to receive and assert, like this https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/test/groovy/ratpack/http/client/HttpProxySpec.groovy#L31

rhart:
This is great, thanks!

ratpack-codahale-metrics/src/main/java/ratpack/codahale/metrics/CodaHaleMetricsModule.java
djjorjinho: I can relocate these "builder" classes to another package.
What do you say about "ratpack.codahale.metrics.internal.builder" or "ratpack.codahale.metrics.builder" ?

djjorjinho:
Added the test assertions and some overall fixes.

ratpack-codahale-metrics/src/main/java/ratpack/codahale/metrics/CodaHaleMetricsModule.java
rhart: I would say just in the `ratpack.codahale.metrics` package for now.

djjorjinho:
I've merged recent changes and relocated the Graphite config class out of the module.

Let me know if there's any further changes you require to the module.

alkemist:
@djjorjinho looks like there are some merge conflicts.

johnrengelman:
Hey @djjorjinho, it seems when you rebased this PR, you may have done it in the wrong direction (i.e. you rebased the upstream master branch onto you local branch). This has caused a number of commits to be rewritten in your pull request.
You should be performing a rebase such that your local branch is rebased onto the current master branch from the ratpack repo. Assuming you have set the `upstream` remote in git to be the ratpack repo, you would do: 
```
$ git checkout my_branch && git fetch upstream master && git rebase upstream/master
```
If you don't know how to fix this locally, I can correct it manually and merge it.

djjorjinho:
Hey @johnrengelman it's my bad, I'm still finding my way around a rebase.
I'll attempt to fix this during this week.

djjorjinho:
I cleaned up my changes. It should be good to go. Let me know if there's anything else you require.

rhart:
Thanks @djjorjinho.  It's still saying there are conflicts but I can sort them while merging.  Will get this in soon as poss.

djjorjinho:
I see, meanwhile there was new changes pushed. Thank you Rus.

PR #734, opened 2015-07-24
View support in JsonRender(er).

### LINKED ISSUE #733: Support for Jackson views
To render an object using specific [views](http://wiki.fasterxml.com/JacksonJsonViews) the current solution is to do:

```
ctx.render(json(myObject, objectMapper.writerWithView(MyView.class)))
```

It would be nice to have `json(Object object, Class<?> view)` in `ratpack.jackson.Jackson` to remove the need to explicitly use the objectMapper and to make it a bit more concise:

```
ctx.render(json(myObject, MyView.class))
```

Raniz85:
Added a viewClass property to JsonRender and implemented support for it
in JsonRenderer.

If the view class is non-null it will be used to create a new
ObjectWriter using that view from the supplied ObjectWriter (either
default or supplied in the JsonRender). If the view class is null
everything will work as before.

This fixes #733

### DIFF
--- a/ratpack-core/src/main/java/ratpack/jackson/Jackson.java
+++ b/ratpack-core/src/main/java/ratpack/jackson/Jackson.java
@@ -288,7 +288,7 @@ public abstract class Jackson {
    * @return a renderable wrapper for the given object
    */
   public static JsonRender json(Object object) {
-    return new DefaultJsonRender(object, null);
+    return new DefaultJsonRender(object, null, null);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


Raniz85:
Not sure why the build fails - it doesn't seem to be related to anything I've done.

alkemist:
Thanks @Raniz-at-Sony .

We are having some strange build issues at Snap CI that we are working on. Your changes passed for me.

Thanks again for the contribution.

PR #761, opened 2015-08-20
MS Word style quotes don't render well in JavaDocs

aaronzirbes:
see http://ratpack.io/manual/current/api/ratpack/server/Service.html

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/761)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/server/Service.java
+++ b/ratpack-core/src/main/java/ratpack/server/Service.java
@@ -126,7 +126,7 @@ import ratpack.exec.Blocking;
  * There is no need to catch promise errors.
  * An error handler for the execution is installed that effectively treats any exceptions as if they were thrown by the method.
  *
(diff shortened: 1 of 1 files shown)


danhyun:
Thanks for this @aaronzirbes smart quotes are the devil

danhyun:
Please add yourself to the contributors list https://github.com/ratpack/ratpack/blob/master/ratpack-manual/src/content/chapters/99-about-the-project.md

danhyun:
It actually looks like something went wrong while creating these docs:

image

alkemist:
We won't be removing the smart quotes. We need to fix the automation that let the docs be built in something other than Unicode.

Thanks for pull request regardless.

danhyun:
This is addressed via this commit https://github.com/ratpack/ratpack/commit/8481348ddab5b643924060b4ad20f7980f4a2058

PR #767, opened 2015-08-27
add the simplest support of custom Guice Injector factory

phigoro:
What allows to use extensions like https://github.com/Netflix/governator.

ratpack.groovy part example:
```groovy
serverConfig {
  props "server.GuiceInjectorFactory": GuiceInjectorFactory.name
}
```
Custom factory example:
```groovy
import com.netflix.governator.guice.LifecycleInjector

class GuiceInjectorFactory extends AbstractGuiceInjectorFactory {
  GuiceInjectorFactory(ServerConfig serverConfig) {
    super(serverConfig)
  }

  @Override
  Injector apply(Module module) throws Exception {
    def builder = LifecycleInjector.builder()
        .inStage(serverConfig.isDevelopment() ? DEVELOPMENT : PRODUCTION)
        .usingBasePackages(GuiceInjectorFactory.package.name)
    if (module)
      builder.withModules(module)
    builder.build().createInjector()
  }
}
```

Initially, I thought of basic registry usage where you optionally put custom factory, but I was not able to quickly get how to implement such approach. As a result I ended up with quick dirty solution based on server config. Please, propose better way if possible.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/767)

### DIFF
--- a/ratpack-guice/src/main/java/ratpack/guice/AbstractGuiceInjectorFactory.java
+++ b/ratpack-guice/src/main/java/ratpack/guice/AbstractGuiceInjectorFactory.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ratpack.guice;
+
+import com.google.inject.Injector;
+import com.google.inject.Module;
+import ratpack.func.Function;
+import ratpack.server.ServerConfig;
+
+public abstract class AbstractGuiceInjectorFactory implements Function<Module, Injector> {
+  protected final ServerConfig serverConfig;
+
+  public AbstractGuiceInjectorFactory(ServerConfig serverConfig) {
+    this.serverConfig = serverConfig;
+  }
+}
--- a/ratpack-guice/src/main/java/ratpack/guice/Guice.java
+++ b/ratpack-guice/src/main/java/ratpack/guice/Guice.java
@@ -33,6 +33,7 @@ import ratpack.server.ServerConfig;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 import static com.google.inject.Guice.createInjector;
 import static ratpack.util.Exceptions.uncheck;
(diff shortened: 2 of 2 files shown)


danveloper:
Hi @phigoro!

First and foremost, thanks for the pull request. I'm not sure how this provides value over just providing the Governator injector as a `parentInjector` when building the registry?

ie:

```java
RatpackServer.start(spec -> spec
  .registry(Guice.registry(LifecycleInjectorBuilder.createInjector(), b -> {
    // BindingsSpec
  }))
  .handlers(c -> ...)
);
```

What's the purpose of driving this through a configuration directive over this approach?

phigoro:
Thanks for quick comment. Yep, parent one looks to be enough. Could you please advise a way to do the same with ratpack.groovy.Groovy.ratpack DSL?

danveloper:
Probably a better option, which fits both use-cases:

```groovy
def governatorRegistry = Guice.registry(Governator.createInjector(..))

ratpack {
  handlers {
    register(governatorRegistry)

    get("foo") { ComponentFromGovernator c ->
      ...
    }
  }
}
```

Likewise:

```java
public static void main(String[] args) {
  Registry governatorRegistry = Guice.registry(Governator.createInjector(..));
  RatpackServer.start(spec -> spec
    .handlers(c -> c
      .register(governatorRegistry)
      .get("foo", ctx -> {
        ComponentFromGovernator c = ctx.get(ComponentFromGovernator.class);
        ...
      })
    )
  );
}
```

danhyun:
Nice one @danveloper :+1:

phigoro:
Thank you, @danveloper. It works like a charm. I think your last comment should hit the documentation, right into "Google Guice integration" section, it's very helpful for newcomers.

danveloper:
Good idea! Thanks again and Happy Ratpack'ing!

phigoro:
I would like to share my experience with parent-child way discussed above.
So, when I go with the following:
```groovy
class Conf {
  String prop1
}

@AutoBindSingleton
class Foo {
  @Inject Conf conf
}

def governatorInjector = ...
ratpack {
  serverConfig {
    props 'conf.properties'
    require '', Conf
  }
  bindings {
    ...
  }
  handlers {
    register(Guice.registry(governatorInjector))
    get {
      context.get(Conf).prop1 // works
      context.get(Foo).conf.prop1 // null, i.e. it's a brand new Conf instance
    }
  }
}
```
Foo is created before Conf, and Conf hits child injector anyway. Also, "ratpack { bindings {...} }" doesn't take place with parent-child way :(, what's intuitively obvious.

Unfortunately, I've fallen back to custom injector factory to get it working as expected.

Do you have any comment or advice around this situation?

alkemist:
@phigoro what's the definition of `governatorInjector`?

phigoro:
@alkemist, thanks for your attention. It was as follows:
```groovy
def governatorInjector = LifecycleInjector.builder()
  .inStage(PRODUCTION)
  .usingBasePackages("com.example")
  .build()
  .createInjector()
```
We might play with lazy init, but it would sound like beans are trying to solve issues of outside stuff and only add complexity and possible another future issues.

alkemist:
@phigoro the problem here just seems to be that `Conf` isn't a singleton.

phigoro:
It looks config is not loaded at all in such situation. There is a test project to ease our discussion: https://github.com/phigoro/ratpack-pr767

PR #779, opened 2015-09-15
updated ratpack-test-groovy to ratpack-groovy-test

solatic:
At some point, there was a name change from ratpack-test-groovy to ratpack-groovy-test, but the name change didn't happen in the documentation on the site, so it's changed here

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/779)


alkemist:
Thanks @solatic .

Would you mind adding yourself to https://github.com/ratpack/ratpack/blob/master/ratpack-manual/src/content/chapters/99-about-the-project.md#L21-21 ?

PR #780, opened 2015-09-15
Support hot reloading in the Groovy Console. Closes #778.

### LINKED ISSUE #778: Unable to reload Ratpack server from Groovy Console
Groovy Version: 2.4.3

Result: Get a `BindException`.

The issue is that the `ThreadLocal` here: https://github.com/ratpack/ratpack/blob/master/ratpack-groovy/src/main/java/ratpack/groovy/internal/StandaloneScriptBacking.java#L46 is returning `null`.

johnrengelman:
The issue stems from the fact that the Groovy Console starts a new
Thread for each invocation of the script in the console. Thus a
ThreadLocal won't contain a reference to the previous launch of the
server. We get around this by making the assumption that if the script
being launched is not backed by a physical file of some kind (as is the
case when running from the console), then we maintain a single JVM wide
static reference to the executed server. In all other cases, we use a
ThreadLocal as we can be fairly certain that the same thread is used to
initiated the original server and any reloads.

The limitiation of this approach is that when running from the Groovy
Console only a single RatpackServer can be launched as each subsequent
launch will shut down the previous server.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/780)

### DIFF
--- a/ratpack-groovy/src/main/java/ratpack/groovy/internal/StandaloneScriptBacking.java
+++ b/ratpack-groovy/src/main/java/ratpack/groovy/internal/StandaloneScriptBacking.java
@@ -30,11 +30,13 @@ import ratpack.util.Exceptions;
 import java.nio.file.Path;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 
 public class StandaloneScriptBacking implements Action<Closure<?>> {
 
   private final static AtomicReference<Action<? super RatpackServer>> CAPTURE_ACTION = new AtomicReference<>(null);
-  private final ThreadLocal<RatpackServer> running = new ThreadLocal<>();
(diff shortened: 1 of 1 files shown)


alkemist:
I see what you were saying now. I'm going to merge this and tweak it slightly. We can simplify this a little 

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ratpack/ratpack/780)

alkemist:
Reviewed 1 of 1 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ratpack/ratpack/780)

PR #785, opened 2015-09-23
Improve ContentCompressor

### LINKED ISSUE #784: Identity Accept-Encoding not always respected
Due to state being stored in the SmartHttpContentCompressor, it's possible for pipelined http requests with different Accept-Encoding headers to cause the identity encoding request to not be respected. 

Looking at further implementations of Compressors state is not normally used across encode and begineEncode, so we shouldn't do that anymore.

beckje01:
IgnorableHttpContentCompressor now acts like the SmartHttpContentCompressor but without state being saved in the class.

The content-encoding header with a value of identity is removed from the response to keep the same behavior as before.

This fixes #784

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/785)


alkemist:
Reviewed 3 of 3 files at r1.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ratpack/ratpack/785)

PR #792, opened 2015-09-25
Provide support for configuring a connection timeout for HttpClient.

brucenunk:
Hi,

I've committed my proposal for supporting connection timeouts for the HttpClient. If somebody could review and let me know if I can be of any more help?

Thanks, James.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/792)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/client/RequestSpec.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/RequestSpec.java
@@ -117,6 +117,12 @@ public interface RequestSpec {
 
   URI getUrl();
 
+  default RequestSpec connectTimeoutSeconds(int seconds) {
+    return connectTimeout(Duration.of(seconds, SECONDS));
+  }
+
+  RequestSpec connectTimeout(Duration duration);
+
   default RequestSpec readTimeoutSeconds(int seconds) {
     return readTimeout(Duration.of(seconds, SECONDS));
   }
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/DelegatingRequestSpec.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/DelegatingRequestSpec.java
@@ -98,6 +98,12 @@ public class DelegatingRequestSpec implements RequestSpec {
   }
 
   @Override
+  public RequestSpec connectTimeout(Duration duration) {
+    delegate.connectTimeout(duration);
+    return null;
+  }
+
+  @Override
   public RequestSpec readTimeout(Duration duration) {
     delegate.readTimeout(duration);
     return this;
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/DelegatingRequestSpec.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/DelegatingRequestSpec.java
@@ -100,7 +100,7 @@ public class DelegatingRequestSpec implements RequestSpec {
   @Override
   public RequestSpec connectTimeout(Duration duration) {
     delegate.connectTimeout(duration);
-    return null;
+    return this;
   }
 
   @Override
--- a/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientSmokeSpec.groovy
+++ b/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientSmokeSpec.groovy
@@ -17,6 +17,7 @@
 package ratpack.http.client
 
 import io.netty.buffer.Unpooled
+import io.netty.channel.ConnectTimeoutException
 import io.netty.handler.codec.http.HttpHeaderNames
 import io.netty.handler.codec.http.HttpHeaderValues
 import io.netty.handler.codec.http.HttpHeaders
@@ -289,6 +290,27 @@ class HttpClientSmokeSpec extends HttpClientSpec {
 
     then:
     response.statusCode == 500
+  }
+
+  def "can set connect timeout"() {
+    setup:
+    def nonRoutableIp = '192.168.0.0'
+
+    when:
+    handlers {
+      get { HttpClient httpClient ->
+        httpClient.get("http://$nonRoutableIp".toURI()) {
+          it.connectTimeout(Duration.ofMillis(1))
+        } onError {
+          render it.class.name
+        } then {
+          render "success"
+        }
+      }
+    }
+
+    then:
+    text == ConnectTimeoutException.name
   }
 
   def "can set read timeout"() {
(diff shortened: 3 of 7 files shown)


danveloper:
Hi @brucenunk! Thank you -- will review!

ratpack-core/src/main/java/ratpack/http/client/internal/DelegatingRequestSpec.java
```diff
@@ -98,6 +98,12 @@ public RequestSpec readTimeoutSeconds(int seconds) {
   }
 
   @Override
+  public RequestSpec connectTimeout(Duration duration) {
+    delegate.connectTimeout(duration);
+    return null;
```
danveloper: `return this`?

ratpack-core/src/test/groovy/ratpack/http/client/HttpClientSmokeSpec.groovy
```diff
@@ -291,6 +292,27 @@ class HttpClientSmokeSpec extends HttpClientSpec {
...
+    setup:
+    def nonRoutableIp = '192.168.0.0'
+
+    when:
+    handlers {
+      get { HttpClient httpClient ->
+        httpClient.get("http://$nonRoutableIp".toURI()) {
+          it.connectTimeout(Duration.ofMillis(20))
+        } onError {
+          render it.class.name
```
danveloper: how about just `render "error"` and assert against that just in case we need to change the underlying behavior in the future?

ratpack-core/src/main/java/ratpack/http/client/internal/DelegatingRequestSpec.java
brucenunk: Oh absolutely. Sorry, my bad.

ratpack-core/src/test/groovy/ratpack/http/client/HttpClientSmokeSpec.groovy
brucenunk: Yeah can do. I took inspiration from the read timeout test. I'm new to Ratpack, Groovy and Spock so you'll have to bear with me.

danveloper: No problem at all :-) I'm still just learning things myself.

erdi:
@brucenunk would you mind also adding yourself as a contributor at https://github.com/ratpack/ratpack/blob/master/ratpack-manual/src/content/chapters/99-about-the-project.md#contributors?

ratpack-core/src/test/groovy/ratpack/http/client/HttpClientSmokeSpec.groovy
brucenunk: Do you want me to make this change Dan? I checked for the specific exception as I wanted to make sure it was a connection timeout not any other exception as I was initially getting tripped up by the read timeout (which I believe defaults to 30 seconds) when I was connecting to a listening address.

danveloper: Ok, that's no problem then. Thanks for clarifying!

danveloper:
image

brucenunk:
Have pushed the requested changes to my fork. Do I need to make another pull request? Sorry this is my first github experience.

danveloper:
Nope! Looks like the build is running right now => https://snap-ci.com/ratpack/ratpack/pull/792 ... We'll see if anyone else from @ratpack/committers wants to review.

ratpack-core/src/main/java/ratpack/http/client/RequestSpec.java
```diff
@@ -117,6 +117,12 @@ default RequestSpec sslContext(Factory<SSLContext> factory) throws Exception {
 
   URI getUrl();
 
+  default RequestSpec connectTimeoutSeconds(int seconds) {
```
alkemist: I think we probably want some kind of sensible default here if this isn't set explicitly. Anyone know of any precedent for the default?

erdi: Are there any defaults in netty?

alkemist: AFAICT, if this isn’t set it falls back to the socket’s SO_TIMEOUT value, whose default is different on different machines.

brucenunk: I definitely agree on a default. I left it optional as I didn't want to change the default behaviour without getting a view from you guys.

Reading some of the Netty code, it appears the default connect timeout is 30000ms (https://github.com/netty/netty/blob/netty-4.1.0.Beta5/transport/src/main/java/io/netty/channel/DefaultChannelConfig.java) which is used by AbstractNioChannel connect (https://github.com/netty/netty/blob/netty-4.1.0.Beta5/transport/src/main/java/io/netty/channel/nio/AbstractNioChannel.java).

Will knock up a quick test to see what's actually going on but I think 30s sounds like a decent default (same as the read timeout in Ratpack from memory?)

brucenunk: I can confirm that the default connect timeout is 30 seconds. Would you like me to make the change to default to that, removing the Optional?

erdi: That's a good idea, go ahead.

brucenunk:
I've committed the change to default connect timeout to 30 seconds. 

As a side, do you think it would be worth binding an instance of something that looked like RequestParams to the registry as "default http client configuration" as things like connect/read timeouts are generally part of SLA's so the same across many requests? The HttpClient could then use this as it's defaults.

erdi:
I think that being able to bind something that would specify defaults to be applied on all `RequestSpec`s used by the default `HttpClient` would be a good idea. I found that such feature would be useful from my own experience. There are ways around it already like for example binding a decorator type around `HttpClient` which applies the defaults because it's a relatively small interface (only 4 methods) and hence pretty easy to decorate. On the other hand it feels like it should be easier to set various default timeouts for all http client calls across the whole app.

I would say that it's a good idea but it's probably worth creating a different issue/PR for it and discussing the implementation in real time with @alkemist in our [community slack channel](http://slack-signup.ratpack.io/) before putting any effort into it.

alkemist:
Thanks @brucenunk

brucenunk:
No worries, happy to be a (small) part of this awesome project :smile:

Issue #777, opened 2015-09-15
No output is present when there is no logger binding


alkemist:
Currently, we don't make it clear enough that a minimum requirement is an slf4j binding. For example, the Groovy quickstart examples don't depend on `slf4j-simple`. We either need to fix this, or provide a better experience when there is no binding.

davidmc24:
Updating the groovy QuickStart examples to use slf4j-simple makes sense to me, and making it clearer in the manual that we maintain the default slf4j behavior when there is no binding: no-op.

erdi:
+1

alkemist:
To finish this off, we need to:

1. Add a page to the manual that talks about logging, and update the no binding warning to point to it.
2. Make sure all of our examples contain a logging binding.

Any takers?

danthegoodman:
@alkemist There was already a logging page, so I just added a link to that in the message. I'm not sure if that section is what you are looking for, but it's a starting point at least.

The Gradle part of the Getting Started section already had a dependency upon slf4j-simple, so I added a `@Grab` statement to the groovy script example.

PR #797, opened 2015-10-03
Deal with missing output when missing a logging implementation

danthegoodman:
Regards issue #777

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/797)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/server/internal/Slf4jNoBindingDetector.java
+++ b/ratpack-core/src/main/java/ratpack/server/internal/Slf4jNoBindingDetector.java
@@ -36,7 +36,8 @@ public abstract class Slf4jNoBindingDetector {
     } catch (ClassNotFoundException e) {
       System.err.println(
         "WARNING: No slf4j logging binding found for Ratpack, there will be no logging output." + System.lineSeparator()
-          + "WARNING: Please add an slf4j binding, such as slf4j-log4j2, to the classpath."
+          + "WARNING: Please add an slf4j binding, such as slf4j-log4j2, to the classpath." + System.lineSeparator()
+          + "WARNING: More info may be found here: http://ratpack.io/manual/current/logging.html"
       );
       return false;
     }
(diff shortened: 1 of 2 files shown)


ratpack-core/src/main/java/ratpack/server/internal/Slf4jNoBindingDetector.java
```diff
@@ -36,7 +36,8 @@ private static boolean init() {
     } catch (ClassNotFoundException e) {
       System.err.println(
         "WARNING: No slf4j logging binding found for Ratpack, there will be no logging output." + System.lineSeparator()
-          + "WARNING: Please add an slf4j binding, such as slf4j-log4j2, to the classpath."
+          + "WARNING: Please add an slf4j binding, such as slf4j-log4j2, to the classpath." + System.lineSeparator()
+          + "WARNING: More info may be found here: http://ratpack.io/manual/current/logging.html"
```
alkemist: I wonder if we should use `RatpackVersion.getVersion()` here instead of current.

alkemist: Decided against this. Going to merge as is.

alkemist:
Thanks @danthegoodman !

ratpack-core/src/main/java/ratpack/server/internal/Slf4jNoBindingDetector.java
erdi: @alkemist would it be better to mention `slf4j-simple` here instead of `slf4j-log4j2`? If a user chooses `slf4j-log4j2` then they will need to also include `log4j2` and configure it but `slf4j-simple` doesn't need any more dependencies and has default config that logs everything from `INFO` up and outputs it to `System.err`.

alkemist: For those reasons yes, but slf4j-simple is blocking so I’d rather not promote it here.

PR #799, opened 2015-10-04
Renderer docs

### LINKED ISSUE #508: Provide prominent documentation of all built in renderers
Not entirely sure the best way to do this. Might be best to keep the docs on `context.render` light, and point to `Renderer`. Then in the class level docs we can have a section on each of the default renderers.

danthegoodman:
Fixes #508 

I compared the set already documented with the set added in the ServerRegistry and found that only two classes were missing. It seems like this is a prominent enough place to document the list. 

I'm not sure if you want to document all of these renderers in the manual, but I saw that the JSON section was stubbed out with a TODO and replaced that one.

I can do the Files section too, but one is going to take more time as I didn't find an example already documented somewhere.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/799)


alkemist:
Many thanks @danthegoodman

PR #803, opened 2015-10-08
Properly call onClose method when connection is closed abruptly by peer.

### LINKED ISSUE #801: WebSocketHandler onClose method is not called when connection is abruptly closed.
Hello.

I noticed that when a websocket connection is closed before a close frame is received, the onClose method of websocket handler is not called.

I have experimented with it in catacumba version of `WebSocketEngine.java` and I found that the solution may be like this:

```java
final DirectChannelAccess directChannelAccess = context.getDirectChannelAccess();
final Channel channel = directChannelAccess.getChannel();
channel.closeFuture().addListener(fu -> {
      try {
        handler.onClose();
      } catch (Exception e) {
        throw uncheck(e);
      }
    });
```

At this line https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/main/java/ratpack/websocket/internal/WebSocketEngine.java#L104

niwinz:
This fixes #801

### DIFF
--- a/ratpack-core/src/main/java/ratpack/websocket/internal/WebSocketEngine.java
+++ b/ratpack-core/src/main/java/ratpack/websocket/internal/WebSocketEngine.java
@@ -102,9 +102,19 @@ public class WebSocketEngine {
           }
         });
 
-        context.getDirectChannelAccess().takeOwnership(msg -> {
+        final DirectChannelAccess directAccessChannel = context.getDirectChannelAccess();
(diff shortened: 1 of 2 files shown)


niwinz:
The tests has failed in something completely unrelated to my pr :confused:

alkemist:
@niwinz thanks for the PR.

It failed because JRuby ate too much memory in our build. I haven't worked out why it's so variable, but I've capped it.

Thanks again.

PR #811, opened 2015-10-12
Add spock words and inherit from Specification

stianl:
As it is stands the ```SiteSmoketest``` sample from the testing chapter compiles, but if you include it in a project it is not executed (e.g. ```gradle test``` will build with no tests run). It think it should be written as a Spock specification or JUnit test case. I made some small changes so it will be picked up and run as a Spock test.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/811)

### DIFF
--- a/ratpack-manual/src/content/chapters/16-testing.md
+++ b/ratpack-manual/src/content/chapters/16-testing.md
@@ -34,24 +34,28 @@ Note below we use @Delegate so we just need to call `get()` in the when block in
 import ratpack.groovy.test.GroovyRatpackMainApplicationUnderTest
 import ratpack.test.http.TestHttpClient
 import ratpack.test.ServerBackedApplicationUnderTest
+import spock.lang.*
 
-class SiteSmokeSpec {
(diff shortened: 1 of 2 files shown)


alkemist:
Hi @stianl ,

Are you sure that this is actually executed now? I don't see how it could be.

While we are here, it would be good to make the demo more self contained, by using `GroovyEmbeddedApp` to create an app that we test.

stianl:
Yes, I am sure it executes now. Note that I am talking about when it is included (typically copy-pasted)  into a project. I've tested again and made a [sample bash script](https://gist.github.com/stianl/95d40c2e0f52cccccda5) that demonstrates it.

alkemist:
It’s not being executed as part of our tests though, which means it could break again without us knowing about it. We’d need to use an alternative fixture that actually executes the test.

alkemist:
@stianl I have reworked that page, and have removed the code listing. This supersedes your changes.

Many thanks for taking the time to submit the contribution and sorry that feedback was late in coming.

Issue #817, opened 2015-10-13
'Connection: Close' directive is not being respected


alkemist:
This should possibly work by closing the connection if the header `Connection: Close` is set on the response during request processing. We could possibly sugar this by adding a `forceConnectionClose()` method to response that just adds this header.

brucenunk:
I think there are two issues here @alkemist. Firstly, DefaultResponseTransmitter setting the "Connection" header regardless of whether it has already been set and secondly, an option to explicitly close the connection once the response has been committed.

The first issue would have zero change to the api, just a check in "pre" to see whether the header has already been set and falling back to existing logic if not (https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/main/java/ratpack/server/internal/DefaultResponseTransmitter.java#L83-L85). Then in "post" check whether "Connection" header is keep-alive otherwise close (https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/main/java/ratpack/server/internal/DefaultResponseTransmitter.java#L234-L236) so I'm quite happy to pick that work up now.

The second issue would involve a change to both Response and ResponseTransmitter apis and I'm not sure whether you want to commit to that for 1.1.0 release?

James.

alkemist:
I'd change that slightly,

I'd set `isKeepAlive` in `pre()` based on the logic you describe. Then, no other methods need to be updated.

BTW, `ResponseTransmitter` is internal so there's no problem changing any of that API.

brucenunk:
OK no worries, will do that. Having a few issues with Intellij at the minute but will get to it this alvo.

alkemist:
Great :)

If we can help with the IDEA stuff holler.

brucenunk:
Hollering ;) I've had it running in Intellij before so no idea what it's boggle is.

I've imported the ratpack.gradle file and getting this:

Error:Can't register given path of type 'SOURCE' because it's out of content root.
Content root: '/home/jamesl/dev/github/ratpack/buildSrc'
Given path: '/home/jamesl/dev/github/ratpack/ratpack-gradle/src/main/groovy'

I've since added it (as far as I can tell), refreshed the project and still the same.

I've also tried ./gradlew idea then importing the ratpack.ipr file but that gives me issues too.
 Any ideas?

brucenunk:
OK I've got the project imported but it's not a gradle-linked project ... I'll deal with it.

[...]

PR #820, opened 2015-10-14
'Connection: Close' directive is not being respected

brucenunk:
Changed pre method in DefaultResponseTransmitter to disable keep-alive if the response headers contain a "Connection: close" header.

https://github.com/ratpack/ratpack/issues/817

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/820)


brucenunk:
Having a few issues with the build. I ran ./gradlew check on my machine before commit and it worked.

Can anybody give me any pointers on what I've done wrong or how to kick off another build please? I can see this in the logs but uncertain what it's on about:

":ratpack-manual:compassCompile FAILED"

Thanks, James.

alkemist:
Thanks @brucenunk

brucenunk:
Thanks for the merge :)

Have been thinking that this change is way to low level but at least it does what I need for now. Definitely think your suggestion of a method on Response is a good idea for explicit closure on commit.

What do you think of a "keep alive" setting on ServerConfig to disable keep alive for the server? Only reason I suggest this is because it seems more likely to me that this would be at a Ratpack instance level rather than per request. 

James.

PR #823, opened 2015-10-19
bumping version to 2.6.1 of asset-pipeline and removing assets prefix…

davydotcom:
… from SiteMain as its not needed anymore

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/823)


alkemist:
Thanks @davydotcom

PR #830, opened 2015-10-29
Specify the default port in the manual

### LINKED ISSUE #771: Add default http://localhost:5050/ to manual
Had to help someone use the default http://localhost:5050/ to see their running app.  I couldn't find a reference to URL  in the manual. Can it  be added? Thinking maybe in both the Quick Start and Launching sections? I usually run the gradle version and see it, but when I ran the groovy script, I didn't see it. might be good to call out in the docs.. if its not there already

Happy to add it myself if needed 

Thanks
-Juan

danthegoodman:
Fixes #771

I added the port to the two sections that I felt like I would look for info if I was just getting started: "Quickstart" and "Launching".

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/830)

### DIFF
--- a/ratpack-manual/src/content/chapters/02-quick-start.md
+++ b/ratpack-manual/src/content/chapters/02-quick-start.md
@@ -36,6 +36,8 @@ You can now start the app by running the following on the command line:
 groovy ratpack.groovy
 ```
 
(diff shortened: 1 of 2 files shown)


alkemist:
More thanks @danthegoodman

Issue #756, opened 2015-08-16
Add option to disable default metric paths in ratpack-codahale-metrics


drmaas:
When this module is installed, by default all unique paths result in a new metric. For large data sets the number of metrics becomes unbounded. If we're sending metrics to external systems like statsd, the network traffic becomes unmanageable. 

I know there is a way of collapsing metrics by group name, but the default metric handler/interceptor still execute. It would be useful to provide an option to disable the default metric handler/interceptor so that we could drop in our own handler/interceptor, while taking advantage of the other features provided by the module.

alkemist:
@rhart seems reasonable to me. WDYT?

rhart:
Sure, makes sense.  I'd still like to understand this a little deeper though.  To see whether there is anything else we can be doing.

@drmaas are you saying that the metric collapsing solves your network traffic issues but you don't want every request to go through the handler?  Even if we allowed requests to be excluded at the handler/interceptor level all requests would still go through the handler, just not all recorded.  But would adding this functionality help?  What would your handler/interceptor do differnet?

drmaas:
It looks like request and execution metrics are recorded. We are only interested in recording request metrics, and we'd like to record timers for the following names:

1) api.response-code.{response code}
2) api.{method}.response-code.{response code}
3) all

It doesn't look like we can achieve that with the current config. If that is indeed the case, we'd like to be able to exclude the injection of ```CodaHaleMetricsModule.HandlerDecoratorProvider```, and instead inject our own handler.

rhart:
Ok, I see.  Will defo put a config option in so you can switch this off.  

Wondering whether we can also allow you to set your own strategy for defining how the metrics are captured, with what we have now as the default.  That way you get the handler decorating and timer capture for free and you can easily provide your own metric format.  Will look into this too.

drmaas:
I think it would be cool to have a config option to allow us to override the default handler with our own - it seems like an elegant solution. It would still be useful to be able to disable it completely for those cases where the user doesn't want to collect any request metrics.

drmaas:
It looks like this is still hardcoded - ideally we could turn off the default handler and blocking interceptor. Let me know if I'm missing something.

```
    bind(RequestTimingHandler.class);
    bind(BlockingExecTimingInterceptor.class);

    Provider<RequestTimingHandler> handlerProvider = getProvider(RequestTimingHandler.class);
[...]

PR #832, opened 2015-10-30
Additional dropwizard metrics customizations

drmaas:
Allow default timing handler and interceptor to be disabled via config, and allow for providing custom implementations of the same.

See issue #756

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/832)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/exec/ExecInterceptor.java
+++ b/ratpack-core/src/main/java/ratpack/exec/ExecInterceptor.java
@@ -22,7 +22,7 @@ import ratpack.func.Block;
 /**
  * Intercepts execution segments of an execution, primarily for traceability and recording metrics.
  * <p>
- * Interceptors present the base registry will be implicitly applied to all executions.
+ * Interceptors present in the base registry will be implicitly applied to all executions.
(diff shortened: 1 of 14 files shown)


rhart:
I'm on hols at the mo, will take a look as soon as I get back

alkemist:
@drmaas sorry for the egregious delay.

Do we need the ability for custom implementations of the interceptors? It adds quite a bit of complexity for not much value.

drmaas:
We'd use it at Target.  Not sure if others would use it, or would want simply disable the defaults. Our main use case is disabling the defaults. Overriding defaults is of secondary import.

Agreed it shouldn't be hard to just add handlers and interceptors independently.

alkemist:
Seems that there's not that much value in the common infrastructure provided by the base classes, that isn't easily done in a custom impl.

Would you mind removing that part of the PR. The rest looks good to merge.

Thanks for your patience.

drmaas:
The latest build failure seems unrelated to this PR, unless I'm missing something.

rhart:
Thanks @drmaas.  I'll do a manual merge on this and sort any build issues out.  

Thanks again for the PR and your patience in getting it merged!

PR #833, opened 2015-10-31
WIP: SSL/TLS reimplemented using Netty facilities instead of JDK

niwinz:
When I started working on http2 support I found that ratpack uses JDK's SSL facilities and I found them a little bit limiting. After read netty documentation / examples / etc I found that Netty ssl facilities are more user friendly than JDK's:

- Them provides simple configurability (needed for http2)
- Does not depends on the obscure and less used jdk keystore format. You can just use the certificate "as is" and private key in pkcs8 format (easy convertible from standard pem format that nginx uses).
- The user api interface looks more familiar (instead of trust/keystore, just a certificate file and private key file).

This PR re-implements the ratpack ssl facilities using Netty api. I don't know if this change is acceptable in ratpack but I think it would be awesome to have because it simplifies the ssl both, in implementation and in the public user api (as I said previously, is more familiar pattern).

Feedback is very welcome!

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/833)


ratpack-core/src/main/java/ratpack/server/ServerConfigBuilder.java
```diff
@@ -178,7 +178,7 @@ default ServerConfigBuilder baseDir(File file) {
    * @see ratpack.ssl.SSLContexts
    * @see ServerConfig#getSslContext()
    */
-  ServerConfigBuilder ssl(SSLContext sslContext);
+  ServerConfigBuilder ssl(SslContext sslContext);
```
alkemist: This is a breaking change.

ratpack-core/src/main/java/ratpack/server/ServerConfig.java
```diff
@@ -153,7 +153,7 @@ static ServerConfig of(Action<? super ServerConfigBuilder> action) throws Except
    * @return The SSL context or <code>null</code> if the application does not use SSL.
    */
   @Nullable
-  SSLContext getSslContext();
+  SslContext getSslContext();
```
alkemist: This is a breaking change.

ratpack-core/src/main/java/ratpack/server/internal/ServerConfigData.java
```diff
@@ -117,11 +117,11 @@ public void setPublicAddress(String publicAddress) throws URISyntaxException {
     this.publicAddress = new URI(publicAddress);
   }
 
-  public SSLContext getSslContext() {
+  public SslContext getSslContext() {
```
alkemist: This is a breaking change.

ratpack-core/src/main/java/ratpack/server/internal/ServerConfigData.java
```diff
@@ -117,11 +117,11 @@ public void setPublicAddress(String publicAddress) throws URISyntaxException {
...
     return sslContext;
   }
 
-  public void setSslContext(SSLContext sslContext) {
+  public void setSslContext(SslContext sslContext) {
```
alkemist: This is a breaking change.

ratpack-core/src/main/java/ratpack/ssl/SSLContexts.java
```diff
@@ -16,180 +16,106 @@
...
 import java.nio.file.Path;
-import java.security.GeneralSecurityException;
-import java.security.KeyStore;
-import java.security.Security;
 
 /**
- * Factory methods for initializing an {@link SSLContext}.
+ * Factory methods for initializing an {@link SslContext}.
  */
 public class SSLContexts {
```
alkemist: This class is public, so can't break.

alkemist:
Hi @niwinz 

Thanks for the PR. We are going to have to work out a way to do this without breaking API, which probably means a lot of overloads.

niwinz:
It is ok for not breaking changes. I can do it without problem (this pr breaks changes for see the final api, and because of that is still WIP). My question here is if this changes are welcome in ratpack. I don't want to do all the hard work of maintain the compatibility do not knowing that.

Later, Is there any deprecation policy defined? Seems a time to have one I think. Django (as example of an other big open source projects has two releases for maintain backward compatibility, more info here: https://docs.djangoproject.com/en/1.8/internals/release-process/#internal-release-deprecation-policy). Is just for know how time we need maintain the old code and leave it clear to the users.

alkemist:
The changes are very welcome. If the new stuff completely overlaps functionally, let's deprecate the old `SSLContext` based methods immediately.

alkemist:
@niwinz finally got around to fixing our CI. Looks like there are compile errors.

niwinz:
@alkemist yes, I have broken the tests in my last commit (is just work in progress). Maybe I should close this PR until is ready for avoid useless builds on the CI?

alkemist:
No problem to leave it open. Was just letting you know that our CI should be working again.

alkemist:
@niwinz Hi. Are you still interested in getting this in?

niwinz:
I'm really want to continue, but I found it completely imposible the last month. I don't have enough time for it :(

I'll close the pr, and maybe in a future I will return to work on it.

alkemist:
No problems, I understand.

PR #834, opened 2015-11-07
Make websocket.send() return ChannelFuture.

achmadns:
I need a way to notify when the text is flushed into websocket client. Could you please incorporate this?

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/834)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/websocket/WebSocket.java
+++ b/ratpack-core/src/main/java/ratpack/websocket/WebSocket.java
@@ -17,6 +17,7 @@
 package ratpack.websocket;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelFuture;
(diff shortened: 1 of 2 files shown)


alkemist:
We can't make this change. It has two problems:

1. It breaks public API
2. It exposes ChannelFuture, which is too level

Two options going forward:

1. Expose new methods to WebSocket that somehow signal that the content has been written
2. Introduce a different interface/usage pattern, perhaps based on Publisher

achmadns:
Sorry, my English is rather...

Is there workaround for now so that I can do something when the text is written to ws client?

alkemist:
The only thing would be to copy/clone the websocket stuff that is there and patch it to your liking. You don't necessarily need to fork Ratpack to do this.

achmadns:
Thanks, I'll try it first.

PR #845, opened 2015-11-13
Use Thymeleaf instead of Jackson as example in Spring docs

dsyer:
Jackson is in core now and there is no need to use a module to
render JSON. A better example is probably the thymeleaf module.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/845)


danhyun:
Thanks Dave

PR #853, opened 2015-12-01
Fix documentation for Asset Pipeline

rahulsom:
1. The version of asset pipeline used depended on an older version of ratpack.
  I've made it use the latest version, which works just fine with current ratpack.

2. There were bits of code that wouldn't compile correctly here that I've fixed.

It is still worth investigating why the tests do not capture those problems.

Thanks a lot to the ratpack developers for walking me through this on slack.
This is one of the best communities when it comes to supporting newbies.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/853)


alkemist:
Many thanks @rahulsom.

Can you please add yourself to this file:  https://github.com/ratpack/ratpack/blob/master/ratpack-manual/src/content/chapters/99-about-the-project.md#contributors

You can just add the commit to this PR.

rahulsom:
@alkemist 

> Can you please add yourself to this file: https://github.com/ratpack/ratpack/blob/master/ratpack-manual/src/content/chapters/99-about-the-project.md#contributors

Done. Thanks!

alkemist:
Thanks @rahulsom !

Issue #828, opened 2015-10-23
Provide method for AUT's to augment/override Guice bindings


johnrengelman:
Related to https://github.com/ratpack/ratpack/issues/680

Need something similar to `MainClassApplicationUnderTest.createOverrides(Registry)` to override/augment Guice bindings in tests.

tednaleid:
+1

Playing with ratpack and creating tests and this was one of the first issues I hit.  Tried using `createOverrides` but guice has already wired everything up so it's too late to inject any mocks that I've created

alkemist:
See https://ratpack.io/manual/1.2.0/api/ratpack/guice/BindingsImposition.html

Issue #743, opened 2015-08-06
MainClassApplicationUnderTest should allow adding extra config


alkemist:
This would be similar to the existing createOverrides() method. Something like:

```
protected addConfig(ConfigDataBuilder builder) {

}
```

This would receive the `ServerConfigBuilder` and users would be expected to _add_ config that overrides the config defined in the main method. The primary use case would be changing things like database handles/credentials etc. to suitable test values.

tomdcc:
I came here looking for an issue to pass args to an app run with MainClassApplicationUnderTest, but it was really just for this use-case. Necessary not just for credentials but testing different behaviour based on different config.

PR #859, opened 2015-12-16
Feature overrides

alkemist:
Proposed solution for #743 and #828.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/859)


johnrengelman:
What about overriding Guice modules/bindings?  This doesn't seem to provide a way to do that. For example, if my app adds a module that has a `@Provide String string()` in it, adding the value to the server/user registry won't override the binding in the Guice injector when it does binding. 

---

Reviewed 26 of 26 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[ratpack-core/src/main/java/ratpack/server/override/UserRegistryOverrides.java, line 22 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/859#-K5eyvMsBnGFgYHsvwYK:-K5eyvMsBnGFgYHsvwYL:1671860046) ([raw file](https://github.com/ratpack/ratpack/blob/4b0a31bdf5724b9dc9d3361715faf28647bfdb8c/ratpack-core/src/main/java/ratpack/server/override/UserRegistryOverrides.java#L22)):*
Javadoc?

---

*[ratpack-test/src/main/java/ratpack/test/MainClassApplicationUnderTest.java, line 43 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/859#-K5f-D8FJ3Rjml30L6mx:-K5f-D8FJ3Rjml30L6my:-1119923845) ([raw file](https://github.com/ratpack/ratpack/blob/4b0a31bdf5724b9dc9d3361715faf28647bfdb8c/ratpack-test/src/main/java/ratpack/test/MainClassApplicationUnderTest.java#L43)):*
Perhaps this is better to be just `protected void override(RegistrySpec spec)`?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ratpack/ratpack/859#-:-K5f-lciTUS4-OianG-p:-141636878)*

alkemist:
We need to add a new override type that the Guice stuff looks for, similarly to how the server config build looks for overrides. As implemented right now, this PR doesn't address the Guice override stuff. It needs to be added, based on this pattern.

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[ratpack-core/src/main/java/ratpack/server/override/UserRegistryOverrides.java, line 22 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/859#-K5eyvMsBnGFgYHsvwYK:-K5hB-yhCjPZAFFiZwtr:-1805134915) ([raw file](https://github.com/ratpack/ratpack/blob/4b0a31bdf5724b9dc9d3361715faf28647bfdb8c/ratpack-core/src/main/java/ratpack/server/override/UserRegistryOverrides.java#L22)):*
Getting there, chipping away at it.

---

*[ratpack-test/src/main/java/ratpack/test/MainClassApplicationUnderTest.java, line 43 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/859#-K5f-D8FJ3Rjml30L6mx:-K5hB4I496OW1gMmOT3-:-240763436) ([raw file](https://github.com/ratpack/ratpack/blob/4b0a31bdf5724b9dc9d3361715faf28647bfdb8c/ratpack-test/src/main/java/ratpack/test/MainClassApplicationUnderTest.java#L43)):*
Not sure, that's less accurate. We aren't overriding the registry spec.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ratpack/ratpack/859#-:-K5hB9MM-lDRqCueqs1x:-392159987)*

alkemist:
The guice stuff is finally in. This is close to complete. Only thing left is docs and more test coverage for the Guice bits.

Please review.

PR #867, opened 2015-12-31
fixed issue with #pac4j authenticator and boundTo on path binding

guspower:
I ran into an issue using the Pac4jAuthenticator with prefixes - it does not correctly form callbackUrls when there is a boundTo value on the path binding.

I have added tests to cover these use cases and modified the implementation accordingly.

G.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/867)

### DIFF
--- a/ratpack-pac4j/src/main/java/ratpack/pac4j/internal/Pac4jAuthenticator.java
+++ b/ratpack-pac4j/src/main/java/ratpack/pac4j/internal/Pac4jAuthenticator.java
@@ -33,6 +33,7 @@ import ratpack.server.PublicAddress;
 import ratpack.session.SessionData;
 import ratpack.util.Types;
 
+import java.net.URI;
 import java.util.List;
 import java.util.Optional;
 
@@ -81,13 +82,28 @@ public class Pac4jAuthenticator implements Handler {
   }
 
   public Clients createClients(Context ctx, PathBinding pathBinding) throws Exception {
-    String boundTo = pathBinding.getBoundTo();
-    PublicAddress publicAddress = ctx.get(PublicAddress.class);
-    String absoluteCallbackUrl = publicAddress.get() + boundTo + "/" + path;
+    String callback = createCallbackUrl(ctx.get(PublicAddress.class), pathBinding, path);
 
+    return new Clients(callback, getClientsFromProvider(ctx, clientsProvider));
+  }
+
+  private String createCallbackUrl(PublicAddress publicAddress, PathBinding pathBinding, String path) {
+    String boundTo = pathBinding.getBoundTo();
+    URI address = publicAddress.get();
+
+    if(!boundTo.isEmpty()) {
+      if (!(address.toString().endsWith("/") || boundTo.startsWith("/"))) {
+        boundTo = "/" + boundTo;
+      }
+    }
+
+    return address + boundTo + "/" + path;
+  }
+
+  @SuppressWarnings("rawtypes")
+  private List<Client> getClientsFromProvider(Context ctx, RatpackPac4j.ClientsProvider clientsProvider) {
     Iterable<? extends Client<?, ?>> result = clientsProvider.get(ctx);
 
-    @SuppressWarnings("rawtypes")
     List<Client> clients;
     if (result instanceof List) {
       clients = Types.cast(result);
@@ -95,7 +111,7 @@ public class Pac4jAuthenticator implements Handler {
       clients = ImmutableList.copyOf(result);
     }
 
-    return new Clients(absoluteCallbackUrl, clients);
+    return clients;
   }
 
   private <C extends Credentials, U extends UserProfile> UserProfile getProfile(WebContext webContext, Client<C, U> client) throws RequiresHttpAction {
--- a/ratpack-pac4j/src/test/groovy/ratpack/pac4j/internal/Pac4jAuthenticatorSpec.groovy
+++ b/ratpack-pac4j/src/test/groovy/ratpack/pac4j/internal/Pac4jAuthenticatorSpec.groovy
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ratpack.pac4j.internal
+
+import ratpack.handling.Context
+import ratpack.pac4j.RatpackPac4j
+import ratpack.path.PathBinding
+import ratpack.server.PublicAddress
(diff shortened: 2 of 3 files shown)


ratpack-pac4j/src/main/java/ratpack/pac4j/internal/Pac4jAuthenticator.java
```diff
@@ -81,13 +82,28 @@ public void handle(Context ctx) throws Exception {
...
+  }
+
+  private String createCallbackUrl(PublicAddress publicAddress, PathBinding pathBinding, String path) {
     String boundTo = pathBinding.getBoundTo();
-    PublicAddress publicAddress = ctx.get(PublicAddress.class);
-    String absoluteCallbackUrl = publicAddress.get() + boundTo + "/" + path;
+    URI address = publicAddress.get();
+
+    if(!boundTo.isEmpty()) {
+      if (!(address.toString().endsWith("/") || boundTo.startsWith("/"))) {
```
erdi: This condition is unnecessary because according to the contracts public address should not end with a `/` and bound path binding should not start with a `/`

ratpack-pac4j/src/main/java/ratpack/pac4j/internal/Pac4jAuthenticator.java
```diff
@@ -81,13 +82,28 @@ public void handle(Context ctx) throws Exception {
...
+        boundTo = "/" + boundTo;
+      }
+    }
+
+    return address + boundTo + "/" + path;
+  }
 
+  @SuppressWarnings("rawtypes")
```
erdi: Is there a way to avoid having this warning at all? What about using a wildcard generic type for `Client` here? Is there an issue with that I'm not seeing?

ratpack-pac4j/src/main/java/ratpack/pac4j/internal/Pac4jAuthenticator.java
guspower: You are correct - the two provided implementations (ConstantPublicAddress and InferringPublicAddress) both reconstruct the URI from scheme, host and port so there would be no trailing slash. However I thought that there was cause to include the check:
1. implementations of PublicAddress can be user provided 
2. the PublicAddress interface returns a URI - which contains path in its public contract. I would be happier if PublicAddress returned an interface that described only the scheme + host + path. It would not be unreasonable for users to add a path to their implementation as it is not obvious that the return type can not contain a path (especially for old skool servlet types who expect to be able to set a web application context).

ratpack-pac4j/src/main/java/ratpack/pac4j/internal/Pac4jAuthenticator.java
guspower: I was trying to minimize the changes to the existing code (which contained the SuppressWarnings check). Wildcarding Client should do the trick.

ratpack-pac4j/src/test/groovy/ratpack/pac4j/internal/Pac4jAuthenticatorSpec.groovy
```diff
@@ -0,0 +1,66 @@
...
+import ratpack.pac4j.RatpackPac4j
+import ratpack.path.PathBinding
+import ratpack.server.PublicAddress
+import ratpack.server.internal.ConstantPublicAddress
+import spock.lang.Specification
+import spock.lang.Unroll
+
+import static ratpack.pac4j.RatpackPac4j.DEFAULT_AUTHENTICATOR_PATH
+
+class Pac4jAuthenticatorSpec extends Specification {
```
erdi: I don't think that this test is really necessary. I think that the convention is that we tend to use integration tests for everything unless there really is no other way and you have provided coverage for your changes by by unrolling the test in `OpenIdRpSpec`.

ratpack-pac4j/src/main/java/ratpack/pac4j/internal/Pac4jAuthenticator.java
guspower: Actually the issue is that Clients has the following constructor:
    
    public Clients(final String callbackUrl, final List<Client> clients) 

so either way it's a suppress or cast. Any ideas about a clean way to get around this?

ratpack-pac4j/src/main/java/ratpack/pac4j/internal/Pac4jAuthenticator.java
erdi: I see, you have a point about the first part of the condition, wrt public address but I'm not sure we should be this defensive here, what do you think @alkemist? On the other hand I'm sure that the second part of the condition can go because  bound path binding will never start with a `/`.

ratpack-pac4j/src/main/java/ratpack/pac4j/internal/Pac4jAuthenticator.java
erdi: That's a shame, let's leave the suppress.

ratpack-pac4j/src/test/groovy/ratpack/pac4j/openid/OpenIdRpSpec.groovy
```diff
@@ -46,6 +48,17 @@ class OpenIdRpSpec extends RatpackGroovyDslSpec {
...
+      prefix 'prefix', {
+        configurePac4j delegate
+      }
+
+      configurePac4j delegate
+    }
+
+  }
+
+  GroovyChain configurePac4j(GroovyChain chain) {
```
erdi: I think we should we use a `GroovyChainAction` implementation here instead of this construct, it will be more idiomatic. I would also add the `noauth` handler only once, in the root of the test application.

ratpack-pac4j/src/test/groovy/ratpack/pac4j/internal/Pac4jAuthenticatorSpec.groovy
guspower: Yeah I deliberated about whether it was useful or not and I know you've said before that you don't like or agree with mocking. However :) I found OpenIdRpSpec was not expressing the cases I wanted to test for clearly - I had hoped to make it easier to understand what was happening with callback url generation (tbh it really should be building a URL rather than a String - but that was a wider change than I wanted to cause). Regarding the mocking, this is stubbing without order or interaction verification - I just wanted to get a PublicAddress and PathBinding in there to exercise the different variants. I realize there is still maintenance downside if the public contract for Context.get(Class), PathBinding.getBoundTo() and RatpackPac4j.ClientsProvider.get(Context) changes, but I figured I'd put it up there for your consideration.

erdi: @alkemist what do you think?

ratpack-pac4j/src/test/groovy/ratpack/pac4j/openid/OpenIdRpSpec.groovy
guspower: Yes - totally agree with GroovyChainAction here. Will update.
The noauth handler is under both root and prefixed paths so needs to be added twice.

guspower: @erdi I've updated the spec to use GroovyChainAction and made the main test a bit more readable.

alkemist:
Thanks @guspower and @erdi. I ended up harvesting parts of this by hand and making some tweaks.

I took the tests as is, as I never say no to free tests :)

PR #876, opened 2016-01-10
Copy response buffer in BlockingHttpClient onto heap instead of leaking the non-heap one

### LINKED ISSUE #875: DefaultTestHttpClient is leaking non-heap memory
The response buffer is retained indefinitely by `BlockingHttpClient` leading to a non-heap memory leak. Given that it's test code it's probably ok to pay the performance penalty of copying the response into a heap byte buffer which can be normally GCed instead of risking running out of memory when testing responses with large bodies.

erdi:
Closes #875.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/876)

### DIFF
--- a/ratpack-test/src/main/java/ratpack/test/internal/BlockingHttpClient.java
+++ b/ratpack-test/src/main/java/ratpack/test/internal/BlockingHttpClient.java
@@ -17,7 +17,6 @@
 package ratpack.test.internal;
 
 import io.netty.buffer.ByteBuf;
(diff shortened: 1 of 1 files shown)


alkemist:
Are you sure that this is necessary, given that the buffer being copied is unpooled?

alkemist:
Ah I see, it's still using a direct buffer. I hadn't consider this. Will need to check for other similar usages.

alkemist:
Going to merge and tweak slightly.

PR #880, opened 2016-01-13
Upgrade to pac4j v1.8

leleuj:
[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/880)

### DIFF
--- a/ratpack-pac4j/src/main/java/ratpack/pac4j/internal/RatpackWebContext.java
+++ b/ratpack-pac4j/src/main/java/ratpack/pac4j/internal/RatpackWebContext.java
@@ -20,31 +20,28 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import io.netty.handler.codec.http.cookie.DefaultCookie;
+import org.pac4j.core.context.Cookie;
 import org.pac4j.core.context.WebContext;
 import org.pac4j.core.exception.RequiresHttpAction;
 import ratpack.form.Form;
 import ratpack.form.internal.DefaultForm;
 import ratpack.form.internal.FormDecoder;
 import ratpack.handling.Context;
-import ratpack.http.HttpMethod;
-import ratpack.http.MediaType;
-import ratpack.http.Request;
-import ratpack.http.TypedData;
+import ratpack.http.*;
 import ratpack.server.PublicAddress;
 import ratpack.session.SessionData;
 import ratpack.util.MultiValueMap;
 
 import java.net.URI;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 
 public class RatpackWebContext implements WebContext {
 
   private final Context context;
   private final SessionData session;
   private final Request request;
+  private final Response response;
   private final Form form;
 
   private String responseContent = "";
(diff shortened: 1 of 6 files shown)


ratpack-pac4j/src/main/java/ratpack/pac4j/internal/RatpackWebContext.java
```diff
@@ -92,23 +109,43 @@ public Object getSessionAttribute(String name) {
...
 
   @Override
   public void setResponseHeader(String name, String value) {
-    context.getResponse().getHeaders().set(name, value);
+    response.getHeaders().set(name, value);
+  }
+
+  @Override
+  public void setResponseCharacterEncoding(String encoding) {
+    response.contentType(encoding);
```
alkemist: I don't think this is right. Will probably have to read the existing content type, and add the encoding to it.

leleuj: Get the content type, look for `;charset=`, add it if missing, change it if it already exists. Sounds good?

alkemist:
Thanks @leleuj - will take a deeper look in the morning.

ratpack-pac4j/src/main/java/ratpack/pac4j/internal/RatpackWebContext.java
alkemist: Something like that, but I'll have to look at what the contract for this method is. It might have to be more intelligent in order to support setting the encoding first and then the type. If that's the case, we'll need to hold on to the encoding in this type and implicitly add it to the type when set.

leleuj: It's not the first time I have troubles to map this method to a framework implementation. I'll take a look on pac4j side to see if I can remove it in favor of the `setResponseContentType` method.

johnrengelman:
@alkemist do we want to actually force this update into the existing module or should we add a separate `ratpack-pac4j1.8` module?

leleuj:
Versions 1.7 and 1.8 of pac4j have breaking changes: the core API has changed. Though, clients mostly remain the same (they are sometimes deprecated when a new one must be used).

The `RatpackPac4j` facade completely hides the changes happening in pac4j. My plan is to left this facade backward compatible and add what's necessary to benefit from the new features of pac4j.

For Ratpack users, it means they can upgrade without worrying. So my recommendation would be to keep the same module.

johnrengelman:
Thanks for the info @leleuj

leleuj:
@alkemist : I just fixed the problem with the `setResponseCharacterEncoding` method: it's now removed from pac4j v1.8.5. The build is broken for now as it depends on the SNAPSHOT: I'll do the release before you merge the pull request of course.

alkemist:
@leleuj so we can be testing changes to the PR, can you please add the following here:

https://github.com/ratpack/ratpack/blob/ced129e78afaa6c1d21b9b7f2ba95a892d8e22bb/ratpack-pac4j/ratpack-pac4j.gradle#L29-L28

```
repositiories {
  maven { url "«repo containing pac4j snapshots»" }
}
```

leleuj:
I just changed it, the build now fails on the tests, the dependency is found.

leleuj:
@alkemist : I just released pac4j v1.8.5 and updated the pull request...

Issue #848, opened 2015-11-20
Crash ratpack if using default configuration from start.spring.io. How to use resource from jar?


lavcraft:
Reproduce
---------------

```bash
curl https://start.spring.io/starter.tgz -d dependencies=ratpack -d type=gradle-project -d baseDir=my-dir | tar -xzvf - && cd my-dir
vi src/main/java/com/example/DemoApplication.java
```

Then edit DemoApplication.java
```java
import ratpack.spring.config.EnableRatpack;

@EnableRatpack
public class DemoApplication {
...
```

Build project and run
```bash
gradle wrapper && ./gradlew build -x test
java -jar build/libs/demo-0.0.1-SNAPSHOT.jar
```

java -jar.. output:
```
Caused by: java.lang.IllegalArgumentException: Filesystem binding must be absolute
        at ratpack.file.internal.DefaultFileSystemBinding.<init>(DefaultFileSystemBinding.java:32)
        at ratpack.file.FileSystemBinding.of(FileSystemBinding.java:52)
        at ratpack.server.internal.DefaultServerConfigBuilder.baseDir(DefaultServerConfigBuilder.java:66)
        at ratpack.spring.config.RatpackConfiguration$ServerConfigConfiguration.ratpackServerConfig(RatpackConfiguration.java:72)
        at ratpack.spring.config.RatpackConfiguration$ServerConfigConfiguration$$EnhancerBySpringCGLIB$$9448c0b4.CGLIB$ratpackServerConfig$0(<generated>)
        at ratpack.spring.config.RatpackConfiguration$ServerConfigConfiguration$$EnhancerBySpringCGLIB$$9448c0b4$$FastClassBySpringCGLIB$$2e50e497.invoke(<generated>)
        at org.springframework.cglib.proxy.MethodProxy.invokeSuper(MethodProxy.java:228)
        at org.springframework.context.annotation.ConfigurationClassEnhancer$BeanMethodInterceptor.intercept(ConfigurationClassEnhancer.java:318)
        at ratpack.spring.config.RatpackConfiguration$ServerConfigConfiguration$$EnhancerBySpringCGLIB$$9448c0b4.ratpackServerConfig(<generated>)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #882, opened 2016-01-15
Create a file system for root of JAR if app is running from archive

dsyer:
If user runs a ratpack app with "java -jar ..." the root of the
classpath is a jar:file:... URI, and this can be converted to a
java.nio.Path, but only with care. This change creates the underlying
Filesystem if necessary and ensures that the JAR URI is in fact valid
before returning it to the Ratpack config.

Users can then use the JAR to store static assets (e.g. in /static
or /public in a Spring Boot app).

Fixes gh-848

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/882)


alkemist:
Thanks @dsyer !

PR #890, opened 2016-01-21
Supports authorizations using pac4j

leleuj:
[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/890)

### DIFF
--- a/ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
+++ b/ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
@@ -231,6 +231,7 @@ public class RatpackPac4j {
    * @param authorizers the authorizers to check authorizations
    * @return a handler
    */
+  @SafeVarargs
   public static Handler security(Class<? extends Client<?, ?>> clientType, Authorizer<UserProfile>... authorizers) {
     return ctx -> RatpackPac4j.login(ctx, clientType).then(userProfile ->
       {
--- a/ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
+++ b/ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
@@ -17,6 +17,7 @@
 package ratpack.pac4j;
 
 import com.google.common.collect.ImmutableList;
+import org.pac4j.core.authorization.Authorizer;
 import org.pac4j.core.client.Client;
 import org.pac4j.core.client.Clients;
 import org.pac4j.core.context.WebContext;
@@ -40,6 +41,7 @@ import ratpack.registry.Registry;
 import ratpack.session.Session;
 import ratpack.util.Types;
 
+import java.util.Arrays;
 import java.util.Optional;
 
 /**
@@ -53,6 +55,9 @@ import java.util.Optional;
  * <p>
  * The {@link #requireAuth(Class)} method provides a handler that acts like a filter, ensuring that the user is authenticated for all requests.
  * This can be used for requiring authentication for all requests starting with a particular request path for example.
+ * <p>
+ * The {@link #security(Class, Authorizer[])} method provides a handler that acts like a filter, ensuring that the user is authenticated and
+ * that authorizations (defined by {@link Authorizer}) are checked for all requests.
  * <p>
  * The {@link #userProfile(Context)}, {@link #login(Context, Class)} and {@link #logout(Context)} methods provide programmatic authentication mechanisms.
  */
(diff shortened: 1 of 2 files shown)


alkemist:
Thanks @leleuj 

Looks like there's a compiler warning (which we treat as error): https://snap-ci.com/ratpack/ratpack/pull/890/logs/defaultPipeline/1/Test

leleuj:
It's fixed

ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
```diff
@@ -173,8 +178,75 @@ public static Handler authenticator(String path, ClientsProvider clientsProvider
...
+   * import static org.junit.Assert.assertEquals;
+   *
+   * public class Example {
+   *   public static void main(String... args) throws Exception {
+   *     EmbeddedApp.of(s -> s
+   *       .registry(Guice.registry(b -> b.module(SessionModule.class)))
+   *       .handlers(c -> c
+   *         .all(RatpackPac4j.authenticator(new IndirectBasicAuthClient(new SimpleTestUsernamePasswordAuthenticator())))
+   *         .prefix("require-authz", a -> a
+   *           .all(RatpackPac4j.security(IndirectBasicAuthClient.class, (ctx, profile) -> { return "user".equals(profile.getId()); }))
```
alkemist: This can be simplified to: 

```
.all(RatpackPac4j.security(IndirectBasicAuthClient.class, (ctx, profile) -> "user".equals(profile.getId())))
```

ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
```diff
@@ -173,8 +178,75 @@ public static Handler authenticator(String path, ClientsProvider clientsProvider
...
+   *   }
+   * }
+   * }</pre>
+   *
+   * @param clientType the client type to use to authenticate with if required
+   * @param authorizers the authorizers to check authorizations
+   * @return a handler
+   */
+  @SafeVarargs
+  public static Handler security(Class<? extends Client<?, ?>> clientType, Authorizer<UserProfile>... authorizers) {
```
alkemist: Should be `Authorizer<? extends UserProfile>`.

alkemist: WDYT about calling this method `authz` or `authorize`? The term `security` is a bit vague and doesn't really describe what it will do.

alkemist: Upon better reading, perhaps just `auth` is better as it is really doing authn and authz.

alkemist:
Quality contribution @leleuj - many thanks. Just a few questions there for you.

ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
leleuj: Yes. Will change.

ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
leleuj: About the `Authorizer<? extends UserProfile>`, I can change that but as the `login` method returns a `UserProfile`, I have an unchecked cast on the user profile: `a.isAuthorized(webContext, (T) userProfile)))`: not a big deal from my point of view. Otherwise, I need to pass the profile type.

What do you think?

leleuj: About the name, I admit `security` is a bit vague, but I'm not really convinced by `auth`: can we think of a verb like `protect` or `secure`?

leleuj:
Thanks for merging. The next and last step is: https://github.com/ratpack/ratpack/issues/896

alkemist:
Regarding the question about the signature, I realised that it's actually a bit broken and fixing it is not going to be easy. See #897. I'm still thinking about solutions.

leleuj:
I think the question about the name of the main method is still open as well: do we keep `security` which is indeed a bit vague or do you think a verb like `secure` or `protect` would be a better solution?

leleuj:
@alkemist : one more thing: I saw you released a 1.2.0-RC1 version and tagged #897 for 1.3.0: it seems to late for #896 to fit in 1.2.0. Though, if we want to rename the `security` method, it must be done for 1.2.0, but I have no problem to delay #896 for 1.3.0.

alkemist:
@leleuj I did rename the method to `secure` for 1.2.0.

Regarding #897, we'll just have to break API in 1.3. I'm willing to do this in this case.
Regarding #896, I think we can just wait for 1.3.

Is that ok with you?

leleuj:
Sounds good to me. If you want me to handle #897, just provide me your guidelines...

PR #903, opened 2016-02-06
Handle Pac4j direct authentication

robfletcher:
Previously RatpackPac4j would always try to force a redirect to establish auth credentials. If the Pac4j auth client extends `DirectClient` that will just throw an exception as they don't support redirects. Here I've short-circuited the redirect if the client is direct.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/903)

### DIFF
--- a/ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
+++ b/ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
@@ -21,7 +21,6 @@ import org.pac4j.core.authorization.Authorizer;
 import org.pac4j.core.client.Client;
 import org.pac4j.core.client.Clients;
 import org.pac4j.core.client.DirectClient;
-import org.pac4j.core.context.WebContext;
 import org.pac4j.core.credentials.Credentials;
 import org.pac4j.core.exception.RequiresHttpAction;
 import org.pac4j.core.exception.TechnicalException;
@@ -201,13 +200,13 @@ public class RatpackPac4j {
    */
   @SafeVarargs
   @SuppressWarnings("varargs")
-  public static Handler requireAuth(Class<? extends Client<?, ?>> clientType, Authorizer<UserProfile>... authorizers) {
-    List<Authorizer<UserProfile>> authorizerList = asList(authorizers);
+  public static <C extends Credentials, U extends UserProfile> Handler requireAuth(Class<? extends Client<C, U>> clientType, Authorizer<? super U>... authorizers) {
+    List<Authorizer<? super U>> authorizerList = asList(authorizers);
     return ctx -> RatpackPac4j.login(ctx, clientType).then(userProfile -> {
       if (authorizerList.isEmpty()) {
         ctx.next(Registry.single(userProfile));
       } else {
-        internalWebContext(ctx).then(webContext -> {
+        RatpackWebContext.from(ctx, false).then(webContext -> {
           if (all(authorizerList, a -> a == null || a.isAuthorized(webContext, userProfile))) {
             ctx.next(Registry.single(userProfile));
           } else {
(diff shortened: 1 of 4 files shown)


ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
```diff
@@ -269,7 +273,15 @@ public static Handler requireAuth(Class<? extends Client<?, ?>> clientType, Auth
...
-      .route(p -> !p.isPresent(), p -> initiateAuthentication(ctx, clientType))
+      .flatMap(p -> {
+        if (!p.isPresent() && DirectClient.class.isAssignableFrom(clientType)) {
+          return performDirectAuthentication(ctx, clientType);
+        } else {
+          return Promise.value(p);
+        }
+      })
+      .route(p -> !p.isPresent() && DirectClient.class.isAssignableFrom(clientType), p -> ctx.clientError(401))
+      .route(p -> !p.isPresent() && IndirectClient.class.isAssignableFrom(clientType), p -> initiateAuthentication(ctx, clientType))
```
robfletcher: I'm not sure if there's a neater way to handle this with `Promise`. Essentially what it's doing is in the `flatMap` attempting to get a `UserProfile` directly from the credentials (token or basic auth header). Because I want to populate the promise I couldn't do it all in `route` so I've got an awkward split of `flatMap` and `route` with a similar conditional.

ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
```diff
@@ -496,4 +508,21 @@ private static void initiateAuthentication(Context ctx, Class<? extends Client<?
...
+      Blocking
+        .get(() -> {
+          Clients clients = ctx.get(Clients.class);
+          Client<?, ?> client = clients.findClient(clientType);
+          return userProfileFromCredentials(client, webContext);
+        })
+    );
+  }
+
+  private static <C extends Credentials, U extends UserProfile> Optional<UserProfile> userProfileFromCredentials(Client<C, U> client, RatpackWebContext webContext) throws RequiresHttpAction {
```
robfletcher: Trying to inline this method just got me tangled up in generic type resolution hell. If I add these same generic types to the `performDirectAuthentication` method I need to propagate them right up to the call site.

ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
```diff
@@ -269,7 +272,20 @@ public static Handler requireAuth(Class<? extends Client<?, ?>> clientType, Auth
...
+          return Promise.value(p);
+        }
+      })
+      .route(p -> !p.isPresent(), p -> {
+        if (supportsRedirect(clientType)) {
+          initiateAuthentication(ctx, clientType);
+        } else {
+          ctx.clientError(401);
+        }
+      })
```
robfletcher: The `flatMap` attempts to get credentials & user profile via direct auth if possible. If we're not doing direct auth or couldn't get valid credentials then we flow into the `route` which will either do redirecting auth as before or just return 401 for authentication clients that don't support redirects.

ratpack-pac4j/src/test/groovy/ratpack/pac4j/AuthenticationSpec.groovy
```diff
@@ -94,4 +127,45 @@ class AuthenticationSpec extends RatpackGroovyDslSpec {
...
+    }
+
+    expect:
+    get('allowed').statusCode == 401
+    requestSpec { r ->
+      r.basicAuth('user', 'user')
+    }.getText('allowed') == 'Hello user'
+
+    and:
+    resetRequest()
```
robfletcher: Pretty easy to miss that this is necessary when doing multiple requests in one feature method. If you don't the following request inherits the basic auth header from the earlier one.

PR #908, opened 2016-02-09
Timed Methods with asynchronous return types should count async completion time

fromanator:
Currently if you annotate a method with Timed and it returns an async return type (like an Observable or Promise) it will not account for the time it takes to return an actual result. Which can be very misleading. This fix will account for the time it takes to complete an Observable or Promise in the Timer metric reported.

[Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/908)

### DIFF
--- a/ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/internal/TimedMethodInterceptor.java
+++ b/ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/internal/TimedMethodInterceptor.java
@@ -25,7 +25,6 @@ import ratpack.exec.Promise;
 
 import javax.inject.Inject;
 import java.lang.reflect.Method;
-import rx.Observable;
 
 /**
  * An implementation of MethodInterceptor that collects {@link com.codahale.metrics.Timer} metrics for any method
@@ -43,9 +42,7 @@ public class TimedMethodInterceptor implements MethodInterceptor {
     Object result;
     try {
       result = invocation.proceed();
-      if (result instanceof Observable<?>) {
-        result = ((Observable) result).finallyDo(timer::stop);
-      } else if (result instanceof Promise<?>) {
+      if (result instanceof Promise<?>) {
         ((Promise<?>) result).result(v -> timer.stop());
       } else {
         timer.stop();
--- a/ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/internal/TimedMethodInterceptor.java
+++ b/ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/internal/TimedMethodInterceptor.java
@@ -21,9 +21,11 @@ import com.codahale.metrics.Timer;
 import com.codahale.metrics.annotation.Timed;
 import org.aopalliance.intercept.MethodInterceptor;
 import org.aopalliance.intercept.MethodInvocation;
+import ratpack.exec.Promise;
 
 import javax.inject.Inject;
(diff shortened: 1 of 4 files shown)


alkemist:
Thanks for the contribution @fromanator .

We might have to solve this another way. We can't have metrics depending on Rx.

An alternative would be to provide a method that times a promise, instead on annotation/interceptor.

fromanator:
From the CI build it looks like it was having the same issue I was having locally (good to know it's not just me): 

:ratpack-manual:compassCompile
Gem::LoadError: Could not find 'rb-inotify' (>= 0.9) among 16 total gem(s)

fromanator:
I am curious as to the concerns of including rxJava as a dependency in metrics.

alkemist:
It's unrelated. Just because someone is using metrics, it doesn't mean that they want to use Rx.

fromanator:
Ah yeah that makes sense. I removed the dependency to rxJava and right now the interceptor just supports Promises.

Sounds like what you suggested to support observables is what we have been doing as a solution locally. Where we have a method that takes a Lambda and handle the cases were the return type is an observable or promise.

drmaas:
If we want to provide our own interceptors, there seem to be 2 ways of doing it - 
1) provide a config optionin the dropwizard module to override the default interceptors
2) manually remove bound interceptors from the dropwizard module and add your own

Thoughts?

ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/internal/TimedMethodInterceptor.java
```diff
@@ -38,12 +39,19 @@
...
     String timerTag = buildTimerTag(invocation.getMethod().getAnnotation(Timed.class), invocation.getMethod());
     final Timer.Context timer = metricRegistry.timer(timerTag).time();
-
+    Object result;
     try {
-      return invocation.proceed();
-    } finally {
+      result = invocation.proceed();
+      if (result instanceof Promise<?>) {
+        ((Promise<?>) result).result(v -> timer.stop());
```
alkemist: This isn't right, it will cause the promise to be subscribed to twice.

Would you mind rebasing and using the [new time() method that I just added to Promise](https://github.com/ratpack/ratpack/commit/ea3b351e5046a077481a9a0dffdce95aac4758ef#diff-d0a7bffcb1db1f94820642af231dbaeeR1120)?

fromanator:
Using the new time() method from a promise to update the Timer duration, instead of using Timer.Context. Also Timer.Context doesn't need to be "stopped" or "closed", it merely provides some static helper methods to update the time.

ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/internal/TimedMethodInterceptor.java
```diff
@@ -37,13 +39,23 @@
...
+    final Timer timer = metricRegistry.timer(timerTag);
+    final Timer.Context timerContext = timer.time();
+    Object result;
     try {
-      return invocation.proceed();
-    } finally {
-      timer.stop();
+      result = invocation.proceed();
+      if (result instanceof Promise<?>) {
+        ((Promise<?>) result).time(duration ->
```
alkemist: The return of the time() method needs to be effectively returned here, so…

```
result = ((Promise<?>) result).time(duration -> …
```

fromanator:
Fixed, added some test cases too.

alkemist:
Thanks @fromanator.

Issue #883, opened 2016-01-18
Allow specifying order of service setup/teardown


donaldquixote:
Ratpack services (referring to the specific interface that provides start/stop lifetime hooks) provide no way for the developer to specify order of invocation. According to documentation, the order of service invocation is dependent on the registry implementation to which the services are registered.

Is there a technical limitation preventing implementation of this capability, or has this just not been requested enough?

Thanks.

alkemist:
The only limitation is in coming up with a decent way to do this. It's an insidious problem, as evidenced by Spring's `Ordered` and the looseness it causes.

In short, if there is need to order the services and and if we can come up with a decent way of expressing desired order… there's no reason it can't be added.

alkemist:
Here's a plan for this:

1. Introduce `@ServiceDependencies` …

```language-java
package ratpack.server;

@Documented
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.TYPE)
@Inherited
public @interface ServiceDependencies {
  Class<? extends Service?>[] value();
}
```

`Service` types can use this annotation to statically declare their dependencies.

2. Introduce `ServiceDependencySpec` …

```language-java
package ratpack.server;

interface ServiceDependencySpec {
  // all service that match the dependents filter, will be marked as being dependent on all services that match the dependencies filter 
  ServiceDependencySpec dependsOn(Predicate<? super Service> dependents, Predicate<? super Service> dependencies)
}
```

On startup, all instances of `Action<ServiceDependencySpec>` (or some sugar wrapper) will be obtained from the user registry and executed to specify service dependencies. There will be no functional difference between dependencies declared this way and by `@ServiceDependencies`. 

This more verbose case is an escape hatch for when dependencies either need to be declared in a more general fashion (e.g. all services of these type depend on all services of this type), or for when the user is not in control of the service type and wishes to impose a dependency.

3. Have Ratpack honour these dependencies by ensuring that dependent services start after and stop before their dependencies.

The general strategy can be described as describing the relationships between services in a directed acyclic graph. Given this information, we can honour any require ordering and also safely start/stop services in parallel that are not dependent which makes startup and shutdown faster.

donaldquixote:
Seems reasonable. It would be very important to explicitly verify the dependency graph is indeed acyclic for the sake of the developer. Also, a sugar spec builder might be nice, something like:

`ServiceDependencySpecBuilder dependents(Class<? extends Service>... dependent);

[...]

PR #935, opened 2016-03-10
Allow ordering services.

alkemist:
Solution for #883.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/935)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/exec/Operation.java
+++ b/ratpack-core/src/main/java/ratpack/exec/Operation.java
@@ -122,14 +122,6 @@ public interface Operation {
     ).operation();
   }
 
-  /**
-   *
-   * @since 1.3
-   */
-  default void result(Action<? super Optional<? extends Throwable>> action) {
-    onError(e -> action.execute(Optional.of(e))).then(() -> action.execute(Optional.<Throwable>empty()));
-  }
-
   static Operation noop() {
     return of(Block.noop());
   }
--- a/ratpack-core/src/main/java/ratpack/exec/Operation.java
+++ b/ratpack-core/src/main/java/ratpack/exec/Operation.java
@@ -122,6 +122,14 @@ public interface Operation {
     ).operation();
   }
 
+  /**
+   *
+   * @since 1.3
+   */
+  default void result(Action<? super Optional<? extends Throwable>> action) {
+    onError(e -> action.execute(Optional.of(e))).then(() -> action.execute(Optional.<Throwable>empty()));
+  }
+
   static Operation noop() {
     return of(Block.noop());
   }
--- a/ratpack-core/src/main/java/ratpack/server/RatpackServer.java
+++ b/ratpack-core/src/main/java/ratpack/server/RatpackServer.java
@@ -55,6 +55,7 @@ import ratpack.server.internal.DefaultRatpackServer;
  * Server objects are thread safe in that every instance method is {@code synchronized}.
  *
  * @see RatpackServerSpec
+ * @see ratpack.service.Service
  * @see #of(Action)
  */
 public interface RatpackServer {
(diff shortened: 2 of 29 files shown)


alkemist:
@ratpack/committers this is ready for review. I think we should improve the cycle error message as a separate change.

I think we should also do a deprecate/move for `ratpack.server.Service` to `ratpack.service.Service` as a separate change.

beckje01:
While I really like the direction this is going I wonder if we need to provide a default way of starting services synchronously  with the old ordering and allow this preferred way of starting services to be opted into. The concern I have is this will be a non obvious breaking change for some applications that are depending on the ordered sync starting of services.

As for the deprecation thats something that will be deprecated then removed in 2.0 right?

---

Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[ratpack-core/src/main/java/ratpack/exec/Operation.java, line 129 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/935#-KCa6BE2pHiEUO-0Tf-W:-KCa6BE2pHiEUO-0Tf-X:-485787105) ([raw file](https://github.com/ratpack/ratpack/blob/3de2ebaed312f8b29be4dd59780077e4b7def246/ratpack-core/src/main/java/ratpack/exec/Operation.java#L129)):*
Where does this get used?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ratpack/ratpack/935#-:-KCa4qybtHaTOykBm0ox:842827776)*

alkemist:
Regarding the breaking change, I think we should just do it and document it clearly in the release notes. It's a clearly superior mechanism and the current behaviour isn't really safe (e.g. adding a new binding can change service ordering in unpredictable ways). So, the benefit here outweighs the cost and adapting for the change will be trivial for most users (e.g. just add some @DependsOn).

As for the deprecations, yeah removal for 2.0.

---

Review status: 0 of 9 files reviewed at latest revision, 2 unresolved discussions.

---

*[ratpack-core/src/main/java/ratpack/exec/Operation.java, line 129 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/935#-KCa6BE2pHiEUO-0Tf-W:-KChkpWhXl-LPb6m4PHn:1855392001) ([raw file](https://github.com/ratpack/ratpack/blob/3de2ebaed312f8b29be4dd59780077e4b7def246/ratpack-core/src/main/java/ratpack/exec/Operation.java#L129)):*
Good catch. I have removed it. It was used in an earlier version of ServicesGraph.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ratpack/ratpack/935#-:-KChk7tOW-oyw1-Ex06I:-487814179)*

beckje01:
The main oddity for this release is it won't throw any errors at boot time from ratpack in particular it just changes the order that things load which can cause more subtle bugs. It seems like we can just document would it be better if we added the new  `ratpack.service.Service`and allowed only it to use the DependsOn async semantics and leave the deprecated class around on the old serial way of doing things. User can opt into the new better way of doing services and by doing that they move off the deprecated class and way of thinking about starting the application.

---

Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ratpack/ratpack/935#-:-KCqQmRwJFb0DJbfR_3G:292413033)*

alkemist:
That's a really good idea. Let's do that.

---

Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ratpack/ratpack/935#-:-KCr6NilK6pgAtCBQiFy:82009129)*

PR #936, opened 2016-03-13
add jwt pac4j test

gschrader:
I was looking for a simple example of how to use the new jwt support and couldn't find one so here one is.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/ratpack/ratpack/936)


gschrader:
I am now using JWTs to secure my [SPA example](https://github.com/gschrader/ratpack-react-boilerplate)

alkemist:
Thanks @gschrader !

Issue #813, opened 2015-10-12
A trailing / in a URL redirect will result in an incorrect url redirect


myott:
A trailing `/` in a URL redirect will result in an incorrect url redirect. 

- This was first noticed on the Ratpack site by navigating to `ratpack.io/manual/` which 301s to `ratpack.io/manual/manual/current`
- I can reproduce this via the following test (I created) in `ratpack.http.client`

```java
  def "Trailing / should be handled"() {
    given:
    otherApp {
      get("foo") {
        redirect(301, "foo/bar")
      }
      get("foo/bar") { render "baz" }
    }

    when:
    handlers {
      get { HttpClient httpClient ->
        httpClient.get(otherAppUrl("foo/")) {
        } then { ReceivedResponse response ->
          render response.body.text
        }
      }
    }

    then:
    text == "baz"
  }
```

This will result in:
```
ERROR r.e.i.DefaultDevelopmentErrorHandler [ratpack-compute-7-3] 404 client error for request to /foo/foo/bar 
```

Is this expected behavior?

danhyun:
As far as this being intentional behavior, it seems as though we should be able to handle this.

For the time being there is a way to get this assertion to pass:

```
def "Trailing / should be handled"() {
    given:
    otherApp {
      get("foo/:part?") { render 'baz' }
    }

    when:
    handlers {
      get { HttpClient httpClient ->
        httpClient.get(otherAppUrl("foo/")).then { ReceivedResponse response ->
            render response.body.text
        }
      }
    }

    then:
    text == "baz"
  }
```

The Path Binding javadoc demonstrates the ability to declare an optional path token http://ratpack.io/manual/snapshot/api/ratpack/handling/Chain.html#path-binding which will match the incoming request.

danhyun:
Actually this scenario can be simplified to demonstrate the issue at hand:

```
class SimpleSpec extends RatpackGroovyDslSpec {
  def "Trailing / should be handled"() {
    given:
    handlers {
      get("foo/:part?") { render 'baz' }
    }

    expect:
    getText('foo') == 'baz'
    getText('foo/') == 'baz'
    getText('foo/baz') == 'baz'
  }
}
```

danhyun:
After carefully rereading the issue I realize this is about redirects ending up at a funky path.

It seems that providing an absolute uri will fix this:

```
class SimpleSpec extends RatpackGroovyDslSpec {
  def "Trailing / should be handled"() {
    given:
    handlers {
[...]

PR #939, opened 2016-03-19
Should use absolute path in manual/current redirect

myott:
This is in reference to https://github.com/ratpack/ratpack/issues/813. 

`manual/` is redirecting to `manual/manual/current` instead of `manual/current`

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/939)


danhyun:
Thanks @myott LGTM

Issue #937, opened 2016-03-14
NcsaRequestLogger enhancements


huxi:
I’d like to enhance `NcsaRequestLogger` a bit:

- change the `DateTimeFormatter` used to always use `Locale.ENGLISH` (​*not*​ Zone).
  This is a trivial change, simply adding `.withLocale(Locale.ENGLISH);` to the private static `FORMATTER`.

  A timestamp like `[14/Mär/2016:15:01:10 +0100]` (the output of the current implementation in Germany) is problematic to parse and also a highly uncommon date format. I consider this one a bug.
- add status-code-based Markers to the logger calls, e.g. `“status-2xx”`, `“status-4xx”` etc.) to enable better filtering options in the logger configuration. I mean general markers for all status codes of a given class (Informational 1xx, Successful 2xx, etc.), not markers for specific status codes. "status-4xx" but not "status-404".
- Use `slf4j` message pattern for formatting instead of the currently used `String.format`.

  This provides better performance, formatting is performed lazily and it also opens the opportunity to filter log-events based on the message pattern. I'm aiming for two distinct patterns: one including the optional id and one without.
- this one might be a bit controversial...

  I’d suggest to log at `WARN` level in case of 4xx and 5xx or if status code is not in the range 100-599...
- write related tests.
  
  Current spec only checks if info is called twice for two requests to Ratpack.

I'd start implementing those changes and preparing a PR after the core devs all had a chance to comment.

danhyun:
As far as log level for error codes, I'm fine with `WARN` but I do know that there are other use cases (hyperic) where WARNs will start cascading error alerts.

I'd be happy to see this implemented

@ratpack/committers thoughts?

alkemist:
Forcing the locale to English makes sense.

Wouldn't the status code markers then make us not be using NCSA format?

Slf4j message pattern… yes please.

Regarding the warning logging, I don't think that should be the default. I could be convinced of an opt in though. 

More tests… yes please.

huxi:
The markers are invisible if not explicitly enabled in the [message pattern](https://logging.apache.org/log4j/2.x/manual/layouts.html#PatternLayout).

In log4j2, a config making use of markers would look like this
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Configuration status="INFO">

  <Appenders>
    <List name="requestList" />
    <List name="serverErrorRequestList">
[...]

PR #940, opened 2016-03-20
Request logger enhancements

huxi:
This implements the changes discussed in #937.

This does not yet expose logging of status code 4xx and 5xx in the public API.

The functionality is there, though. Everything is tested and tests broken by this change have been fixed.

I'd suggest to merge this already and discuss how it should be exposed in the public API separately.


michaelschlies:
@ldaley, @beckje01 Is there value in the implementation of this, after this much time? Or more specifically, is there value in the approach being taken? This would be easy to update to fix conflicts and re-PR.

PR #958, opened 2016-04-15
Configure Methods that can have Bodies

beckje01:
Allow the user to configure which HttpMethods could have bodies.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/958)


alkemist:
@beckje01 why do you need this? I'd like to try and avoid the need for this config option if possible.

beckje01:
@alkemist The list of methods we mark as allowing a body isn't set in the HTTP spec, DELETE requests can have a body but it isn't widely used. I wanted to allow developers to pick what request types could have bodies so we don't impose limits but we also don't want to treat every request as having a body so I thought a config option would be best to leave it up to the developers.

In my personal usecase, we have a Ratpack 1.1.0 app that expects a body on DELETE requests that works fine but all upgrades of ratpack will not work due to the change to body parsing.

alkemist:
Closing this in favour of removing the need to know which kinds of requests can have bodies.

/cc @beckje01

PR #964, opened 2016-04-25
Execute promises in parallel and return promise for all results

drmaas:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/964)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/exec/Promise.java
+++ b/ratpack-core/src/main/java/ratpack/exec/Promise.java
@@ -266,7 +266,7 @@ public interface Promise<T> {
    * @see #async(Upstream)
    * @see Execution#fork()
    */
-  static <T> Promise<Set<Result<T>>> forkEach(Set<Promise<T>> promises) {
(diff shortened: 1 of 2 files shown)


drmaas:
Some background on this:

We frequently have use cases where we have to issue multiple concurrent IO requests, say to Cassandra. We want to not only issue the requests as non-blocking and/or in parallel on separate IO threads, but process the responses in parallel as well on ratpack execution threads.

After significant amounts of performance testing, this code scales and performs better than any alternatives, like using RxJava.

alkemist:
Hi @drmaas ,

This is great stuff and we are very grateful for the contribution. I'm a little swamped right now, but will do my best to get to this very soon.

drmaas:
Is there a scenario where we want to do the same thing, but without forking new executions for each promise? I'm thinking that would be useful if someone wanted to ensure execution order. Let me know your thoughts.

alkemist:
@drmaas I'm going to close this in favour of #977. It will support the same use case, but provides a space to grow the options for this kind of thing. Please check it out and provide any feedback.

PR #974, opened 2016-05-01
Route Errors during DefaultPromise.then to the execution error handler

### LINKED ISSUE #973: Errors raised in DefaultPromise.then are routed to the promise error handler
We're using code in our handlers like:
```java
promise.onError(throwable -> // do stuff)
    .then(foo -> ctx.next(Registry.single(Foo.class, foo))
```

When a handler further down the chain throws an exception this is wrapped by DefaultContext into a HandlerException (which is an Error).

Since the implementation of Downstream.success in DefaultPromise.then only catches subtypes of Exception, these errors (and any other Error thrown by the action given to Promise.then) will propagate back upstream and end up being routed to the promise's own error handler. This is different from the handling given to Exception/RuntimeException, which are routed to the execution's error handler.

Currently we're working around this with:

```java
promise.onError(throwable -> // do stuff)
    .next(foo -> ctx.next(Registry.single(Foo.class, foo))
    .then(Action.noop())
```

which gives the behaviour we want, but I propose that the error handling behaviour of Promise.then actions should be made consistent with respect to Error, Exception and RuntimeException (by routing them all to the execution).

mkavanagh:
Fixes #973 (errors in Promise.then action are routed to the promise's error handler instead of the execution's error handler) by catching Throwable instead of Exception in the implementation of Downstream.success.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/974)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/exec/internal/DefaultPromise.java
+++ b/ratpack-core/src/main/java/ratpack/exec/internal/DefaultPromise.java
@@ -41,7 +41,7 @@ public class DefaultPromise<T> implements Promise<T> {
       public void success(T value) {
         try {
           then.execute(value);
-        } catch (Exception e) {
+        } catch (Throwable e) {
           throwError(e);
         }
       }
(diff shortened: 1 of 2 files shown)


alkemist:
Thank you very much!

Issue #840, opened 2015-11-08
Provide methods/constructs for concurrent data processing based on promises


alkemist:

danveloper:
Is there any implication to parallelism here?

alkemist:
Yeah, the processing would be concurrent so if there are multiple cores then it would be parallel.

PR #977, opened 2016-05-09
Support for (parallel) batch processing (#840).

alkemist:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/977)


beckje01:
Reviewed 11 of 11 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[ratpack-core/src/main/java/ratpack/exec/util/Batch.java, line 41 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KHKhXE70Pyo_P29j4ir:-KHKhXE70Pyo_P29j4is:-228308009) ([raw file](https://github.com/ratpack/ratpack/blob/8ea80b2814cb5d7c576b7f3648605512ae1be8e7/ratpack-core/src/main/java/ratpack/exec/util/Batch.java#L41)):*
> ```Java
> 
> /**
>  * A mechanism for processing batches of jobs as {@link Promise promises}.
> ```

We don't really talk about what a job is in this context anywhere in our docs should we explain a bit more what job means here or is there a less overloaded term we could use?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/977)*

ratpack-core/src/main/java/ratpack/exec/util/Batch.java
```diff
@@ -0,0 +1,365 @@
...
+    List<Promise<T>> jobs = Lists.newArrayList(this.jobs);
+    if (jobs.isEmpty()) {
+      return Promise.value(Collections.emptyList());
+    }
+
+    return execInit == null ? serialYieldAll(jobs) : parallelYieldAll(jobs, execInit);
+  }
+
+  /**
+   * Processes the batch jobs, return {@link Result} objects for each.
```
drmaas: This method is actually returning Promise for a list of promised values, not Result objects.

ratpack-core/src/main/java/ratpack/exec/util/Batch.java
```diff
@@ -0,0 +1,365 @@
...
+   * <p>
+   * Every job will be processed, regardless of any failure.
+   * <p>
+   * The order of the returned list corresponds to the original job order.
+   * That is, it is guaranteed that the 2nd item in the list was the 2nd job specified.
+   * It does not reflect the order in which jobs completed.
+   *
+   * @return a promise for the result of each job
+   */
+  public Promise<List<? extends ExecResult<T>>> yieldAll() {
```
drmaas: I think it would be more clear to rename this 'yieldResults'.

ratpack-core/src/main/java/ratpack/exec/util/Batch.java
```diff
@@ -0,0 +1,365 @@
...
+   * If processing in parallel, subsequent errors may also occur.
+   * Such errors will be {@link Throwable#addSuppressed(Throwable)} suppressed by the first error.
+   * <p>
+   * The order of the returned list corresponds to the original job order.
+   * That is, it is guaranteed that the 2nd item in the list was the 2nd job specified.
+   * It does not reflect the order in which jobs completed.
+   *
+   * @return a promise for the result of each job
+   */
+  public Promise<List<? extends T>> yield() {
```
drmaas: I think it would be more clear to rename this 'yieldValues'.

ratpack-core/src/main/java/ratpack/exec/util/Batch.java
alkemist: I don't understand. It does return results.

ratpack-core/src/main/java/ratpack/exec/util/Batch.java
alkemist: I'm not sure. The main reason why you would use this or the *all variant is whether you want to stop at first error or not.

drmaas:
General comment - blocks where you have new BiAction(…) { … }  could be replaced with lambda expressions - is there a reason to leave them in the more verbose form?

ratpack-core/src/main/java/ratpack/exec/util/Batch.java
```diff
@@ -0,0 +1,365 @@
...
+   * @param jobs the batch jobs
+   * @param <T> the type of item produced by each job
+   * @return a new batch
+   */
+  @SafeVarargs
+  @SuppressWarnings("varargs")
+  public static <T> Batch<T> of(Promise<T>... jobs) {
+    return new Batch<>(Arrays.asList(jobs), null);
+  }
+
```
drmaas: You could also add static methods Batch.parallel(jobs) and Batch.parallel(jobs, execInit) as a convenience method to Batch.of(jobs).parallel() and Batch.of(jobs).parallel(execInit). This might make the parallelization features more visible.

ratpack-core/src/main/java/ratpack/exec/util/Batch.java
drmaas: Ok. My confusion is in that yield and yieldAll also have different return types. What if someone wants a promise of a list of execution results, but they want to stop on the first error?

alkemist:
It's recursive, in a sense, and therefore needs to pass a this reference. Lambdas can't self reference.

ratpack-core/src/main/java/ratpack/exec/util/Batch.java
alkemist: It's hard to think of a reason why they would want to do this.

alkemist:
Review status: 8 of 11 files reviewed at latest revision, 5 unresolved discussions.

---

*[ratpack-core/src/main/java/ratpack/exec/util/Batch.java, line 41 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KHKhXE70Pyo_P29j4ir:-KHMNRLaASsvEnbB_M5G:99730712) ([raw file](https://github.com/ratpack/ratpack/blob/8ea80b2814cb5d7c576b7f3648605512ae1be8e7/ratpack-core/src/main/java/ratpack/exec/util/Batch.java#L41)):*
<details><summary><i>Previously, beckje01 (Jeff Beck) wrote…</i></summary>
> We don't really talk about what a job is in this context anywhere in our docs should we explain a bit more what job means here or is there a less overloaded term we could use?</details>
It's a good point. I'm considering backing away from the “job” terminology entirely. If we don't do that though, we should elaborate on job == promise.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/977)*

ratpack-core/src/main/java/ratpack/exec/util/Batch.java
drmaas: To clarify, I was referring to the comment header.

beckje01:
Review status: 8 of 11 files reviewed at latest revision, 5 unresolved discussions.

---

*[ratpack-core/src/main/java/ratpack/exec/util/Batch.java, line 41 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KHKhXE70Pyo_P29j4ir:-KHQXXQEGQQtZf-SB0-o:-1213988504) ([raw file](https://github.com/ratpack/ratpack/blob/8ea80b2814cb5d7c576b7f3648605512ae1be8e7/ratpack-core/src/main/java/ratpack/exec/util/Batch.java#L41)):*
<details><summary><i>Previously, alkemist (Luke Daley) wrote…</i></summary>
> It's a good point. I'm considering backing away from the “job” terminology entirely. If we don't do that though, we should elaborate on job == promise.</details>
Yeah I agree we should back away from "job" terminology it think it has too much baggage.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/977)*

alkemist:
Review status: 6 of 11 files reviewed at latest revision, 5 unresolved discussions.

---

*[ratpack-core/src/main/java/ratpack/exec/util/Batch.java, line 41 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KHKhXE70Pyo_P29j4ir:-KHQtuEzU5v0cth0wtdG:-260459341) ([raw file](https://github.com/ratpack/ratpack/blob/8ea80b2814cb5d7c576b7f3648605512ae1be8e7/ratpack-core/src/main/java/ratpack/exec/util/Batch.java#L41)):*
<details><summary><i>Previously, beckje01 (Jeff Beck) wrote…</i></summary>
> Yeah I agree we should back away from "job" terminology it think it has too much baggage.</details>
This is done.

---

*[ratpack-core/src/main/java/ratpack/exec/util/Batch.java, line 121 \[r2\]](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KHIwrHBSFZmYCk585Oe-r2-121:-KHQtwIf0sGgSuNxpntm:1933912048) ([raw file](https://github.com/ratpack/ratpack/blob/ba55436cab8592fac2eb5e45c5e4982ccf9694a2/ratpack-core/src/main/java/ratpack/exec/util/Batch.java#L121)):*
<details><summary><i>Previously, drmaas (Dan Maas) wrote…</i></summary>
> You could also add static methods Batch.parallel(jobs) and Batch.parallel(jobs, execInit) as a convenience method to Batch.of(jobs).parallel() and Batch.of(jobs).parallel(execInit). This might make the parallelization features more visible.</details>
I'm reluctant to do this because we'd have to add 4 extra methods (two source types * two parallel()  variants).

---

*[ratpack-core/src/main/java/ratpack/exec/util/Batch.java, line 167 \[r2\]](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KHIwrHBSFZmYCk585Oe-r2-167:-KHQuBra7iAzVVMGQtvL:1758359711) ([raw file](https://github.com/ratpack/ratpack/blob/ba55436cab8592fac2eb5e45c5e4982ccf9694a2/ratpack-core/src/main/java/ratpack/exec/util/Batch.java#L167)):*
<details><summary><i>Previously, drmaas (Dan Maas) wrote…</i></summary>
> To clarify, I was referring to the comment header - it says return Result for each, but returns <? extends T> objects for each.</details>
Fixed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/977)*

alkemist:
Review status: 6 of 11 files reviewed at latest revision, 5 unresolved discussions.

---

*[ratpack-core/src/main/java/ratpack/exec/util/Batch.java, line 157 \[r2\]](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KHIwrHBSFZmYCk585Oe-r2-157:-KHQuGSvFQScXl2oBqm5:2004766686) ([raw file](https://github.com/ratpack/ratpack/blob/ba55436cab8592fac2eb5e45c5e4982ccf9694a2/ratpack-core/src/main/java/ratpack/exec/util/Batch.java#L157)):*
<details><summary><i>Previously, drmaas (Dan Maas) wrote…</i></summary>
> I think it would be more clear to rename this 'yieldResults'.</details>
As mentioned elsewhere, the result aspect is not really the point. The difference between yield() and yieldAll() is about whether it stops on error or not.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/977)*

beckje01:
Reviewed 2 of 3 files at r2, 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/977)*

alkemist:
Review status: 9 of 11 files reviewed at latest revision, 5 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KHXbYbeoiiZG4a1sbpH:-KHXbYbeoiiZG4a1sbpI:1521978239) (no related file):*
This is now ready for review.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/977)*

alkemist:
I'm still not feeling comfortable with this API for some reason. I can't quite put my finger on it.

I'm wondering if we should have `SerialBatch` and `ParallelBatch` as different types.

danveloper:
What would be the differentiation between a `SerialBatch` and a `Publisher`?

alkemist:
Serial batch will process the promises 1 at a time, and emit one at a time. A parallel publisher will process in parallel, but emit 1 at a time (in the order that they complete).

johnrengelman:
I think we should break the types into `SerialBatch` and `ParallelBatch`. The code is  bit unwieldily with the serial/parallel versions of private methods.

<details><summary><i>Previously, alkemist (Luke Daley) wrote…</i></summary>
> Serial batch will process the promises 1 at a time, and emit one at a time. A parallel publisher will process in parallel, but emit 1 at a time (in the order that they complete).
</details>


---

Review status: 9 of 11 files reviewed at latest revision, 5 unresolved discussions.

---

*[ratpack-core/src/main/java/ratpack/exec/Promise.java, line 425 \[r5\]](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KI87987QOAmF8uDwAao:-KI87987QOAmF8uDwAap:778706030) ([raw file](https://github.com/ratpack/ratpack/blob/b947d3fd487464ac68ceb69ed13b74f76b1d4f73/ratpack-core/src/main/java/ratpack/exec/Promise.java#L425)):*
> ```Java
>       }
>     });
>   }
> ```

Would be nice to have a `batch()` method on `Promise`.
Within it would check if `T` is instance of `Collection` and pass that directly to `Batch.of()`, if not, it can wrap the single value in a  `List` (so a list of size 1) and pass that to `Batch.of()`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/977#-:-KI82auR-wqJXlI8aw71:618916029)*

alkemist:
I'm not concerned about the internal implementation complexity. I'll take that if it means a better user API.

johnrengelman:
@alkemist so you're trying to decide if the user has to decide on serial vs parallel upfront when creating the `Batch` instead of being implicitly serial and calling `parallel()`?

alkemist:
@johnrengelman not necessarily. We may have `Batch` and `ParallelBatch`, and the `parallel()` methods return a `ParallelBatch`. 

I have two concerns with the current API:

1. The parallelism might not be obvious enough
2. It prevents us from having parallel or serial specific features

1) can be solved via documentation. 2) is the bigger concern.

alkemist:
If we took 2) seriously from my previous list, we'd end up with `abstract Batch`, `SerialBatch`, `ParallelBatch`.

johnrengelman:
@alkemist ok, #2 was the same idea in my head but poorly textualized. I think for #2 it would make the most sense to break the classes as you've said.

alkemist:
@johnrengelman made that change.

johnrengelman:
Reviewed 1 of 3 files at r4, 1 of 1 files at r5, 8 of 9 files at r6.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[ratpack-core/src/main/java/ratpack/exec/batch/ParallelBatch.java, line 125 \[r6\]](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KIAWQdIx3rqftnZDQmI:-KIAWQdJDzf2qDxIEJEC:546041692) ([raw file](https://github.com/ratpack/ratpack/blob/3da04872c08d62afdf9caac34f0105eeea587800/ratpack-core/src/main/java/ratpack/exec/batch/ParallelBatch.java#L125)):*
> ```Java
>    */
>   @Override
>   TransformablePublisher<T> publisher();
> ```

Would it make sense to have a `serial()` method on this (and a `parallel()` method on SerialBatch to convert between the two?

---

*[ratpack-core/src/main/java/ratpack/exec/batch/SerialBatch.java, line 28 \[r6\]](https://reviewable.io:443/reviews/ratpack/ratpack/977#-KIAW_TAXHWHP4fAe0R0:-KIAW_TAXHWHP4fAe0R1:-648857552) ([raw file](https://github.com/ratpack/ratpack/blob/3da04872c08d62afdf9caac34f0105eeea587800/ratpack-core/src/main/java/ratpack/exec/batch/SerialBatch.java#L28)):*
> ```Java
> 
> /**
>  * A batch of promises to be processed, in parallel.
> ```

"in serial"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/977)*

johnrengelman:
Reviewed 6 of 12 files at r1, 2 of 3 files at r3, 1 of 9 files at r6.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/977)*

PR #982, opened 2016-05-13
update static files documentation for latest release, closes #981

### LINKED ISSUE #981: serving static files
Hi there, I am trying to serve static files, but documentation seems outdated, as it suggests to call [files("assets")](https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/main/java/ratpack/handling/Chain.java#L311) but there doesn't seem to be a method ***find*** for String with location.
Even if call ***files()*** with 0 arguments, i get ***ratpack.file.BaseDirRequiredException: no base dir set for application***
I can submit a PR with the updates, but would like to clarify the simplest way to serve statics with the default ***Base Dir*** set

jxs:
should the description be also updated?

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/982)

### DIFF
--- a/ratpack-manual/src/content/chapters/18-static-assets.md
+++ b/ratpack-manual/src/content/chapters/18-static-assets.md
@@ -10,8 +10,9 @@ import ratpack.server.RatpackServer;
 public class Main {
   public static void main(String... args) throws Exception {
     RatpackServer.start(server -> server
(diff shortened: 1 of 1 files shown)


alkemist:
Thanks for the contribution. I've done a more thorough reworking, and made the examples tested (39790b081d6084890d157845ff23fa9f9aa95b4a).

Issue #914, opened 2016-02-18
Advertise a logical location in the request handling control flow via path binding description


alkemist:
## Background

We've had a long standing deficiency in that there's no convenient way to identify “what” handled a request. A large contributor to this is that our handling is composed of functions (that have a weak notion of identity), where other MVC type frameworks use classes/methods (that have a strong notion of identity). This affects diagnotics, including error messages, New Relic integration, metric integration and more.

## Proposed solution

We can use the string representation of the cumulative path binding to indicate location.

Given:

```
handlers {
  prefix("foo/:b") {
    get("baz") { 
      render "ok"
    }
  }
}
```

A request to `/foo/bar/baz` would be handled by the “location” `foo/:b/baz`.
A request to `/foo/foo/baz` would be handled by the “location” `foo/:b/baz`.

## Implementation

We already make the given `PathBinding` at any point in time available via the execution internally (see `PathHandler`). We just need to have `PathBinding` give a string representation of itself, based on the originating `PathBinder`.

We could take roughly the following steps:

1. Have TokenPathBinder accept an extra string with is effectively its source path spec (e.g. `foo/:b`)
2. Make the string form the canonical form of a path binder

For 2, this would involve:

1. Having DefaultPathBinderBuilder build a path string
2. Have DefaultPathBinderBuilder.parse() turn the path string into the executable form

This is in contrast to the way things work now where the executable form is canonical, and DefaultPathBinderBuilder.parse() uses DefaultPathBinderBuilder.build().

tylerbenson:
This is great!  I would also suggest providing some information as to the HTTP Method if relevant, since the same path could go to different handlers depending on GET/POST/etc.

brandonwirick:
In addition to a "canonical" form, which I imagine would be valuable for collision detection and handler naming, I think a "searchable" form would be valuable. If I just see `foo/:b/baz`, I don't know that `foo/:b` is the prefix, which is a string literal I could search for in code. A "searchable" form might just be the "canonical" form with whitespace added, like `foo/:b / baz`.

alkemist:
@brandonwirick I'm not sure here. That's a pretty heavy cognitive price to pay.

What do you have in mind where this would be so important?

brandonwirick:
[...]

PR #984, opened 2016-05-17
Support for describing the path binding specification (#914).

alkemist:
This can be used as a kind of logical description of the type of request.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/984)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/path/internal/DefaultPathBinding.java
+++ b/ratpack-core/src/main/java/ratpack/path/internal/DefaultPathBinding.java
@@ -24,34 +24,14 @@ public class DefaultPathBinding implements PathBinding {
 
   private final String binding;
   private final String pastBinding;
+  private final String description;
 
   private final PathTokens tokens;
   private final PathTokens allTokens;
 
-  @Override
-  public boolean equals(Object o) {
-    if (this == o) {
-      return true;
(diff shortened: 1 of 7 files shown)


alkemist:
This is ready for review. 

The only thing I'm unsure about is the name of the `PathBinding.getSpec()` method. The idea is that this string describes what the cumulative binders bind to (i.e. the specification).

ratpack-core/src/main/java/ratpack/path/internal/DefaultPathBinding.java
```diff
@@ -66,18 +48,65 @@ public DefaultPathBinding(String binding, ImmutableMap<String, String> tokens, P
...
+    } else {
+      Map<String, String> merge = new LinkedHashMap<>(parentTokens);
+      merge.putAll(thisTokens);
+      return DefaultPathTokens.of(ImmutableMap.copyOf(merge));
+    }
+  }
+
+  @Override
+  public String getSpec() {
+    return description;
```
tylerbenson: Seems weird that the method is called spec, but the field is called description.

alkemist: Fixed.

PR #987, opened 2016-05-19
Fixes #986 - provides ability to incorporate configuration from existing objects

### LINKED ISSUE #986: Add sugar for using objects as config sources
Via the `ConfigDataBuilder`, we should provide the means to map an object as a config source. This will essentially be sugar for mapping the object's tree to a path in the configuration.

danveloper:
Closes #986

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/987)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/config/ConfigDataBuilder.java
+++ b/ratpack-core/src/main/java/ratpack/config/ConfigDataBuilder.java
@@ -353,6 +353,15 @@ public interface ConfigDataBuilder {
   ConfigDataBuilder yaml(URL url);
 
   /**
+   * Adds the object's fields at the given path as a configuration source.
+   *
+   * @param path the configuration path the object's fields should be mapped on to
+   * @param object the object from which to derive the configuration fields
+   * @return {@code this}
+   */
+  ConfigDataBuilder object(String path, Object object);
+
+  /**
    * Sets the error all that will be used for added configuration sources.
    * The error all only applies to configuration sources added after this method is called; it is not applied retroactively.
    *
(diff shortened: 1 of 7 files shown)


ratpack-core/src/main/java/ratpack/config/ConfigDataBuilder.java
```diff
@@ -353,6 +353,15 @@ default ConfigDataBuilder args(String prefix, String separator, String[] args) {
   ConfigDataBuilder yaml(URL url);
 
   /**
+   * Adds the object's fields at the given path as a configuration source.
+   *
+   * @param path the configuration path the object's fields should be mapped on to
+   * @param object the object from which to derive the configuration fields
+   * @return {@code this}
```
alkemist: Needs `@since 1.4`.

ratpack-core/src/test/groovy/ratpack/config/ConfigDataCreateSpec.groovy
```diff
@@ -140,11 +142,24 @@ class ConfigDataCreateSpec extends RatpackGroovyDslSpec {
     ex.cause instanceof NoSuchFileException
   }
 
+  def "can load config from an object"() {
```
alkemist: Would be good to have test with a nested path, and merging values.

ratpack-core/src/main/java/ratpack/config/ConfigDataBuilder.java
danveloper: added now ⬇️

ratpack-core/src/test/groovy/ratpack/config/ConfigDataCreateSpec.groovy
danveloper: added ⬇️

PR #990, opened 2016-05-25
Kotlin wrapper initial version

jaguililla:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/990)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,4 @@
 org.gradle.jvmargs=-Dfile.encoding=UTF-8
+
+dokkaVersion=0.9.+
+kotlinVersion=1.0.+
--- a/gradle/kotlinModule.gradle
+++ b/gradle/kotlinModule.gradle
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2013 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: 'kotlin'
+apply plugin: 'jacoco'
+apply plugin: 'org.jetbrains.dokka'
+
+ext.getProperty = { String name, String value ->
+    project.hasProperty (name)? project.property (name) : value
+}
+
+ext.testngVersion = getProperty ('testngVersion', '6.9.+')
+ext.mockitoVersion = getProperty ('mockitoVersion', '1.10.+')
+
+ext.filterPatterns = getProperty ('filterPatterns', '*.properties, *.txt, *.xml')
+ext.rootBuildDir = rootProject.buildDir
+
+repositories {
+    jcenter ()
+    mavenCentral ()
+    mavenLocal ()
+}
+
+dependencies {
+    compile group: 'org.jetbrains.kotlin', name: 'kotlin-reflect', version: kotlinVersion
+
+    testCompile group: 'org.jetbrains.kotlin', name: 'kotlin-test', version: kotlinVersion
+    testCompile group: 'org.testng', name: 'testng', version: testngVersion
+    testCompile group: 'org.mockito', name: 'mockito-core', version: mockitoVersion
+}
+
+processResources {
+    filterPatterns.split(',').each {pattern ->
+        filesMatching ('**/' + pattern.trim ()) {
+            filter {
+                it
+                    .replace ('${projectVersion}', project.version)
+                    .replace ('${projectName}', project.name)
+                    .replace ('${bootstrapVersion}', bootstrapVersion)
+            }
+        }
+    }
+}
+
+test {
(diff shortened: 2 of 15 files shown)


danveloper:
Is it possible in the tests to use the `TestHttpClient` instead?

gradle.properties
```diff
@@ -1 +1,4 @@
 org.gradle.jvmargs=-Dfile.encoding=UTF-8
+
+dokkaVersion=0.9.+
+kotlinVersion=1.0.+
```
johnrengelman: You need to specify extact versions. No dynamic selectors, please.
Also, we already have a standard of defining library versions for the project here: https://github.com/ratpack/ratpack/blob/master/gradle/dependencies.gradle

gradle/kotlinModule.gradle
```diff
@@ -0,0 +1,71 @@
...
+
+ext.getProperty = { String name, String value ->
+    project.hasProperty (name)? project.property (name) : value
+}
+
+ext.testngVersion = getProperty ('testngVersion', '6.9.+')
+ext.mockitoVersion = getProperty ('mockitoVersion', '1.10.+')
+
+ext.filterPatterns = getProperty ('filterPatterns', '*.properties, *.txt, *.xml')
+ext.rootBuildDir = rootProject.buildDir
```
johnrengelman: None of these `ext` methods/properties should be used. Everything should be statically configured.

gradle/kotlinModule.gradle
```diff
@@ -0,0 +1,71 @@
...
+
+repositories {
+    jcenter ()
+    mavenCentral ()
+    mavenLocal ()
+}
```
johnrengelman: These are not needed. They are pre-configured by the root project (https://github.com/ratpack/ratpack/blob/master/ratpack.gradle#L89-L92)

gradle/kotlinModule.gradle
```diff
@@ -0,0 +1,71 @@
...
+
+dependencies {
+    compile group: 'org.jetbrains.kotlin', name: 'kotlin-reflect', version: kotlinVersion
+
+    testCompile group: 'org.jetbrains.kotlin', name: 'kotlin-test', version: kotlinVersion
+    testCompile group: 'org.testng', name: 'testng', version: testngVersion
+    testCompile group: 'org.mockito', name: 'mockito-core', version: mockitoVersion
```
johnrengelman: The pattern this project is to use the string version, not the map version for dependencies

gradle/kotlinModule.gradle
```diff
@@ -0,0 +1,71 @@
...
+        filesMatching ('**/' + pattern.trim ()) {
+            filter {
+                it
+                    .replace ('${projectVersion}', project.version)
+                    .replace ('${projectName}', project.name)
+                    .replace ('${bootstrapVersion}', bootstrapVersion)
+            }
+        }
+    }
+}
```
johnrengelman: I don't see any resources added by in this PR that would need processing. What's this for?

gradle/kotlinModule.gradle
```diff
@@ -0,0 +1,71 @@
...
+
+test {
+    useTestNG ()
```
johnrengelman: We use Spock in this project, is that possible?

ratpack-kotlin/src/test/kotlin/ratpack/kotlin/HttpClient.kt
```diff
@@ -0,0 +1,61 @@
+package ratpack.kotlin
+
+import okhttp3.*
+import java.net.URL
+
+class HttpClient (val base: URL = URL ("http://localhost:5050")) {
```
johnrengelman: Is the existing `TestHttpClient` from `ratpack-test` not sufficient?

johnrengelman:
@jaguililla we are talking within @ratpack/committers how we want to provide/maintain this. So, you don't need to address my comments yet until we decide on that (of course, if you want to, go for it).

alkemist:
Hi @jaguililla 

Thanks for the contribution. I'm not entirely sure that we want to take on having this in the core. We don't really have the expertise to maintain it.

Do you think that it would be possibly for this to exist outside of the distribution? Does it need access to any unstable internals?

gradle/kotlinModule.gradle
```diff
@@ -0,0 +1,71 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+apply plugin: 'kotlin'
+apply plugin: 'jacoco'
+apply plugin: 'org.jetbrains.dokka'
+
+ext.getProperty = { String name, String value ->
+    project.hasProperty (name)? project.property (name) : value
```
otrosien: since gradle 2.13 there is project.findProperty()  https://docs.gradle.org/current/release-notes#project.findproperty%28%29-method-for-safer-property-lookups

jaguililla:
Hi, this code is already in another project. But I think it fits better here. Where it is more visible and more people can contribute. I understand your concerns, and I will maintain the module happily if you agree to accept the code.

The code as it is misses some documentation, but it is stable.

Please consider my proposal of maintaining it (it is not a problem as I'll support that code in production applications anyway).

I will await your reply prior to make the PR changes.

Thanks!

gradle/kotlinModule.gradle
jaguililla: Thanks for the info I didn't knew!

jaguililla:
I will stop using Ratpack by now (I will go back to servlets for a number of reasons), so I won't be able to maintain the code. I'll close the PR. If you want to support Kotlin eventually I would be glad to help (if time allows) cheers!

Issue #465, opened 2014-10-07
Implement Connection Pooling/KeepAlive in the DefaultHttpClient


dannwebster:
Ratpack's DefaultHttpClient for managing outbound HTTP requests currently builds up and tears down every HTTP. This issue addresses this by correctly managing HTTP 1.1 KeepAlive connections, and providing a pooling mechanism for accessing those connections.

chbatey:
Was any work done on this @beckje01 @dannwebster ?

tednaleid:
With the addition of [`ChannelPool`](http://netty.io/news/2015/05/07/4-0-28-Final.html) in netty 4 which we've now got in RP 1.2 that seems like potentially a good way to implement this.

alkemist:
I didn’t know this had landed. Excellent!

Thanks for the pointer.

PR #994, opened 2016-05-27
Adding pooled http client

drmaas:
This drastically improves performance for applications making high volumes of http calls. Typically we see a 2-5x improvement in throughput and average response time when using the pooled http client vs. the default non-pooled client.

Addresses #465

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/994)


alkemist:
Super cool. Will get to it as soon as I can.

alkemist:
We can't merge this as is as there's too much duplication. 

We should rework all of the current impl to work with ChannelPool etc. but to effectively use a pool (etc.) that does no pooling. Then we just have to change the pooling strategy.

I'd also like to flip this API around a bit and introduce …

```
interface HttpClientPool extends AutoCloseable {
    static HttpClientPool of(Action<? super HttpClientPoolSpec> action) {
       // …
    }

    HttpClient getClient()

     // just here as an example, doesn't need to be in this changeset
     int getActiveConnections()
}

interface HttpClientPoolBuilder {
   
}
```

This gives us more space to enrich the pooling and allow users to take control over the connection lifecycle (i.e. close the pool). If someone wants to implicitly pool by default, they can always override the default binding for `HttpClient`.

drmaas:
Hey @alkemist 

We are willing to continue working on this, if you feel the above comments are enough as we work to come up with more unified approach. Another option is that we could turn it over to a core member if they have a better idea on what needs to happen.

The easiest thing to do would be to replace the non-pooling http client with the pooled client code, with a ChannelPoolMap of size 1, and a ChannelPool implementation that does no pooling. Then we could tie in the interfaces you mentioned above.

Let me know how you suggest we proceed with this.

alkemist:
@drmaas our first goal should be to just reduce all of the duplication. Replacing the non-pooling impl with a pooled impl of size 1 as you say sounds like the way to do this. That makes sense as the next step. If you're willing to do that, it would be really great.

_After_ that, we'll look at introducing the `HttpClientPool` abstraction. It might be easiest if I do that, but we can discuss that when we get to it.

Thanks again for all this work. It's really great.

drmaas:
We'll take the next step to remove code duplication, and commit it back for review.

drmaas:
We have a non-pooling refactor almost ready, but are seeing a slight performance degradation vs the current client implementation, of about 5-10% reduction in throughput at high loads (> 200 RPS). I can only assume this is due to some aspect of the pooling setup that we haven't been able to identify yet.

drmaas:
@alkemist 

We have the client replace complete, but there are 7 failing tests that we will need to address.

At this point we don't know why the tests are failing, and will be looking into it. If you or anyone has thoughts on that, it would be much appreciated.

ratpack-core/src/main/java/ratpack/http/client/HttpClient.java
```diff
@@ -106,8 +112,20 @@ static HttpClient httpClient(ServerConfig serverConfig, Registry registry) {
    * @param maxContentLengthBytes The max content length of a response to support.
    * @return An instance of a HttpClient
    */
-  static HttpClient httpClient(ByteBufAllocator byteBufAllocator, int maxContentLengthBytes) {
-    return new DefaultHttpClient(byteBufAllocator, maxContentLengthBytes);
+  static HttpClient httpClient(ByteBufAllocator byteBufAllocator, int maxContentLengthBytes, ExecController execController) {
```
johnrengelman: This is a breaking binary change. You'll need to introduce this as a new method and then modify the existing method with 2 args to call the new method with a sane default for the new parameter.

ratpack-core/src/main/java/ratpack/http/client/internal/ContentAggregatingRequestAction.java
```diff
@@ -30,35 +35,53 @@
...
 
-class ContentAggregatingRequestAction extends RequestActionSupport<ReceivedResponse> {
+public class ContentAggregatingRequestAction extends RequestActionSupport<ReceivedResponse> {
 
-  private final int maxContentLengthBytes;
+  public static final Logger LOGGER = LoggerFactory.getLogger(ContentAggregatingRequestAction.class);
 
-  public ContentAggregatingRequestAction(Action<? super RequestSpec> requestConfigurer, URI uri, Execution execution, ByteBufAllocator byteBufAllocator, int maxContentLengthBytes, int redirectCount) {
-    super(requestConfigurer, uri, execution, byteBufAllocator, redirectCount);
+  private int maxContentLengthBytes;
```
johnrengelman: why did this become not `final`?

ratpack-core/src/main/java/ratpack/http/client/internal/ContentAggregatingRequestAction.java
```diff
@@ -30,35 +35,53 @@
 
 import java.net.URI;
 
-class ContentAggregatingRequestAction extends RequestActionSupport<ReceivedResponse> {
+public class ContentAggregatingRequestAction extends RequestActionSupport<ReceivedResponse> {
 
-  private final int maxContentLengthBytes;
+  public static final Logger LOGGER = LoggerFactory.getLogger(ContentAggregatingRequestAction.class);
```
johnrengelman: Unused - can be removed (along with imports).

ratpack-core/src/main/java/ratpack/http/client/internal/ContentStreamingRequestAction.java
```diff
@@ -47,54 +52,71 @@
...
   protected void addResponseHandlers(ChannelPipeline p, Downstream<? super StreamedResponse> downstream) {
-    p.addLast("httpResponseHandler", new SimpleChannelInboundHandler<HttpResponse>(false) {
+    addHandler(p, "httpResponseHandler", new SimpleChannelInboundHandler<HttpResponse>(false) {
+      private boolean isKeepAlive;
+
       @Override
-      public void channelRead0(ChannelHandlerContext ctx, HttpResponse msg) throws Exception {
+      protected void channelRead0(ChannelHandlerContext ctx, HttpResponse msg) throws Exception {
+        isKeepAlive = HttpUtil.isKeepAlive(msg);
+        msg.headers().forEach(h -> LOGGER.info(h.getKey() + "=" + h.getValue()));
```
johnrengelman: is this logging just for debugging while you're working on this?

johnrengelman:
@drmaas the PR is being run through CI...it's currently failing with `codenarc` errors in `ratpack-core`. This was the latest run - https://snap-ci.com/ratpack/ratpack/pull/994/logs/defaultPipeline/6/Test

ratpack-core/src/main/java/ratpack/http/client/internal/ContentStreamingRequestAction.java
drmaas: yes, forgot to remove it.

ratpack-core/src/main/java/ratpack/http/client/HttpClient.java
drmaas: What would a sane default for ExecController be?

We didn't want to make this change, honestly, but we were unable to find a way to to build a client without knowing the ExecController at bind time, since the  Netty connection pools a bootstrap that depends on getting a handle to the event loop group.

drmaas:
AFAICT the tests are failing due to changes not related to this PR. Let me know if there is something I can address here.

ratpack-retrofit2/src/test/groovy/ratpack/retrofit/RatpackRetrofitSpec.groovy
```diff
@@ -17,7 +17,9 @@
 package ratpack.retrofit
 
 import io.netty.buffer.UnpooledByteBufAllocator
+import ratpack.exec.ExecController
```
beckje01: This is the import causing the codenarc issue it unused, just remove it.

drmaas:
Failing test is ContentStreamingRequestActionSpec > client channel is closed when response is not subscribed to[1]. I can't get the test to fail when run locally. Not sure why it fails in snap-ci.

alkemist:
@drmaas is this ready to hand over as far as your concerned?

drmaas:
@alkemist yes it is. I have the  commits split up to represent different authors.

alkemist:
Superseded by #1020.

Issue #1000, opened 2016-06-05
Add convenience methods to `Operation`


danhyun:
While working with the API, I've found that I've been writing

```
Blocking.op {
  // thing
}.next(Blocking.op {
  // next blocking op
}).then {
}
```

Adding `Operation#blockingNext(Block)` as sugar

PR #1001, opened 2016-06-05
Add convenience method to `Operation` for performing Blocking calls

danhyun:
Addresses #1000

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1001)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/exec/Operation.java
+++ b/ratpack-core/src/main/java/ratpack/exec/Operation.java
@@ -107,6 +107,18 @@ public interface Operation {
     return next(Operation.of(operation));
   }
 
+  /**
+   * Executes the given block as an operation, on a blocking thread.
+   * <p>
+   *
+   * @param operation a block of code to be executed, on a blocking thread
+   * @return an operation
+   * @since 1.4
+   */
+  default Operation blockingNext(Block operation) {
+    return next(Blocking.op(operation));
+  }
+
   default <O> O to(Function<? super Operation, ? extends O> function) throws Exception {
     return function.apply(this);
   }
--- a/ratpack-core/src/main/java/ratpack/exec/Operation.java
+++ b/ratpack-core/src/main/java/ratpack/exec/Operation.java
@@ -109,7 +109,6 @@ public interface Operation {
 
   /**
    * Executes the given block as an operation, on a blocking thread.
-   * <p>
    *
    * @param operation a block of code to be executed, on a blocking thread
    * @return an operation
--- a/ratpack-core/src/test/groovy/ratpack/exec/BlockingSpec.groovy
+++ b/ratpack-core/src/test/groovy/ratpack/exec/BlockingSpec.groovy
@@ -273,6 +273,38 @@ class BlockingSpec extends RatpackGroovyDslSpec {
     events == ["compute", "blocking", "inner compute", "inner blocking"]
   }
 
+  def "ensure the appropriate thread pool processes the segment"() {
+    given:
+    def events = []
+
+    when:
+    handlers {
+      all {
+        Blocking.op {
+          events << Thread.currentThread().name
+        } next {
+          events << Thread.currentThread().name
+        } next(Operation.of {
+          events << Thread.currentThread().name
+        }) next(Blocking.op {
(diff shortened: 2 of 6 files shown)


ratpack-core/src/main/java/ratpack/exec/Operation.java
```diff
@@ -107,6 +107,18 @@ default Operation next(Block operation) {
     return next(Operation.of(operation));
   }
 
+  /**
+   * Executes the given block as an operation, on a blocking thread.
+   * <p>
```
alkemist: rogue <p>

ratpack-core/src/main/java/ratpack/exec/Operation.java
```diff
@@ -107,6 +107,18 @@ default Operation next(Block operation) {
...
+   *
+   * @param operation a block of code to be executed, on a blocking thread
+   * @return an operation
+   * @since 1.4
+   */
+  default Operation blockingNext(Block operation) {
+    return next(Blocking.op(operation));
+  }
+
```
johnrengelman: Should we also have `blockingNext(Operation)` since `next` can take `Block` or `Operation`?

johnrengelman: Or would that be redundant?

danhyun:
*[ratpack-core/src/main/java/ratpack/exec/Operation.java, line 121 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/1001#-KJXEV__rl5r7z3iDOWQ-r1-121:-KJkgV51uNWvydood9oz:-2044985316) ([raw file](https://github.com/ratpack/ratpack/blob/adbfb4a9062458978863d959da0f3a0cfb3da969/ratpack-core/src/main/java/ratpack/exec/Operation.java#L121)):*
<details><summary><i>Previously, johnrengelman (John Engelman) wrote…</i></summary>
> Or would that be redundant?
</details>
I was thinking about this but then I realized that by the time the `Promise` or `Operation` is created it's already determined where it will run either by `Promise.async()`, `Promise.value()`, `Blocking.get()`, `Blocking.op()`, etc

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1001)*

johnrengelman:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[ratpack-core/src/main/java/ratpack/exec/Operation.java, line 121 \[r1\]](https://reviewable.io:443/reviews/ratpack/ratpack/1001#-KJXEV__rl5r7z3iDOWQ-r1-121:-KJkkN6bhfER4yCPv7o7:1314880371) ([raw file](https://github.com/ratpack/ratpack/blob/adbfb4a9062458978863d959da0f3a0cfb3da969/ratpack-core/src/main/java/ratpack/exec/Operation.java#L121)):*
<details><summary><i>Previously, danhyun (Danny) wrote…</i></summary>
> I was thinking about this but then I realized that by the time the `Promise` or `Operation` is created it's already determined where it will run either by `Promise.async()`, `Promise.value()`, `Blocking.get()`, `Blocking.op()`, etc
</details>
Agreed. Not necessary

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1001)*

ratpack-core/src/test/groovy/ratpack/exec/BlockingSpec.groovy
```diff
@@ -273,6 +273,37 @@ class BlockingSpec extends RatpackGroovyDslSpec {
...
 
+  def "ensure the appropriate thread pool processes the segment"() {
+    given:
+    def events = []
+
+    when:
+    handlers {
+      all {
+        Blocking.op {
+          events << Thread.currentThread().name.split('-')[1]
```
alkemist: Another way is to use `Execution.isBlockingThread()`

ratpack-core/src/main/java/ratpack/exec/Operation.java
alkemist: yes we should, @danhyun can you please add.

PR #1018, opened 2016-07-07
Allow for 3rd party reporter extension

fromanator:
Change from the default protected to public so users can utilize these classes when they want to implement a 3rd party reporter (for example statsd). This makes it a lot easier to extend the existing module and follow the same patterns.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1018)


alkemist:
Many thanks for the contribution.

PR #1019, opened 2016-07-13
Allow all requests to have a body.

beckje01:
We check the headers to see if a request should have a body then set up
the request body accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1019)


alkemist:
What? no test? :)

PR #1022, opened 2016-07-18
#998 added shortcut for getting header values

### LINKED ISSUE #998: Provide Context.header(String) as shortcut for context.getRequest().getHeader()

kyleboon:
Closes #998 

This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1022)


ratpack-core/src/main/java/ratpack/handling/Context.java
```diff
@@ -339,6 +339,22 @@
   void render(Object object) throws NoSuchRendererException;
 
   /**
+   * Returns the header value with the specified header name.
```
johnrengelman: This should state that it is a Request header.

Issue #1023, opened 2016-07-18
Add ability to control content negotiation when client doesn't provide Accept header


davidmc24:
There was some discussion about the handling of Accept headers in byContent blocks in #ratpack-dev yesterday.  In particular, some people were confused by the current behavior when there is no Accept handler.  While the documentation doesn't currently call it out, the current behavior in that case is to use the first provided block.

We confirmed that this was indeed both intentional and in compliance with the HTTP spec.  However, we realized that there currently isn't a way for the user to override this behavior.  Thus, we came up with a proposed change, below:
- Add "unspecified(Block)" and "unspecified(String)" similar to noMatch (to ByContentSpec), but to allow controlling the no accept header case
- Leave the default behavior if unspecified isn't called to use the first defined block
- Leave noMatch as-is.
- Update the documentation to clearly specify this behavior, and the existing default behavior.

This adds the ability to control the case, while retaining backwards compatibility.

davidmc24:
An additional case which wasn't specified particularly clearly is the handling of an "Accept" header with no value (empty).  Our interpretation is that it is valid (as per RFC 7231) in this case to either send a 406 error or disregard the header field and send whatever content we feel like.  Our feeling was that it's slightly more user-friendly to return the server's preferred content type rather than an error.  Additionally, that was already the behavior, thus preserving backwards compatibility if we don't change it.  Thus, the empty "Accept" header case will use the "unspecified" handler, not the "noMatch" handler.

Related discussion:
https://stackoverflow.com/questions/12130910/how-to-interpret-empty-http-accept-header

PR #1033, opened 2016-07-25
Add ability to control content negotiation when client doesn't provide Accept header

davidmc24:
See #1023 for further details.

---

This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1033)


ratpack-core/src/main/java/ratpack/handling/ByContentSpec.java
```diff
@@ -84,9 +89,26 @@
...
    */
   ByContentSpec noMatch(String mimeType);
 
+  /**
+   * Specifies that the given handler should be used if the client did not provide a usable "Accept" header in the request.
+   *
+   * @param block the code to invoke if no usable "Accept" header is present in the request.
+   * @return this
+   */
+  ByContentSpec unspecified(Block block);
```
alkemist: Missing `@since`

ratpack-core/src/main/java/ratpack/handling/ByContentSpec.java
```diff
@@ -84,9 +89,26 @@
...
+
+  /**
+   * Specifies that the handler for the specified content type should be used if the client did not provide a usable "Accept" header in the request.
+   * Effectively, this treats the request as if the user requested the specified MIME type.
+   *
+   * @param mimeType the MIME type to use as a fallback if no type is requested
+   * @return this
+   */
+  ByContentSpec unspecified(String mimeType);
```
alkemist: Missing `@since`

ratpack-core/src/main/java/ratpack/handling/Context.java
```diff
@@ -274,9 +274,14 @@
    *   }
    * }
    * }</pre>
-   * If there is no type registered, or if the client does not accept any of the given types, by default a {@code 406} will be issued with {@link Context#clientError(int)}.
+   * If there is no type registered, or if the client does not accept any of the given types, the "noMatch" handler will be used.
```
alkemist: Perhaps we should push all of this doc to `ByContentSpec` and just link to it.

ratpack-core/src/main/java/ratpack/handling/internal/DefaultByContentSpec.java
```diff
@@ -90,7 +96,26 @@ public ByContentSpec noMatch(String mimeType) {
...
+  public ByContentSpec unspecified(Block handler) {
+    unspecifiedHandler = ctx -> handler.execute();
+    return this;
+  }
+
+  @Override
+  public ByContentSpec unspecified(String mimeType) {
+    unspecifiedHandler = ctx -> {
+      ctx.getResponse().contentType(mimeType);
+      blocks.get(mimeType).execute();
```
alkemist: What happens if there's no block for the content type?

ratpack-core/src/main/java/ratpack/handling/internal/ContentNegotiationHandler.java
```diff
@@ -45,20 +46,18 @@ public void handle(Context context) throws Exception {
...
-    }
-
-    if (Strings.isNullOrEmpty(winner)) {
-      noMatchHandler.handle(context);
+    if (Strings.isNullOrEmpty(acceptHeader)) {
+      unspecifiedHandler.handle(context);
     } else {
-      context.getResponse().contentType(winner);
-      blocks.get(winner).execute();
+      List<String> types = blocks.keySet().stream().collect(Collectors2.toListReversed());
```
alkemist: Can be replaced with ImmutableList.of().reverse().

ratpack-core/src/main/java/ratpack/handling/Context.java
davidmc24: Done.

ratpack-core/src/main/java/ratpack/handling/internal/DefaultByContentSpec.java
davidmc24: Previously, it failed with a NullPointerException.  Now, it calls the ServerErrorHandler with a meaningful error message (and has tests to verify that behavior).  I think that it's the right thing, in this case, to fail at runtime, because doing the check at configuration-time would require the block to be registered before the noMatch/unspecified, and I don't think we should require those declarations to be in a particular order.

ratpack-core/src/main/java/ratpack/handling/internal/ContentNegotiationHandler.java
davidmc24: Done.

PR #1041, opened 2016-08-01
Adding method to install handlers and chains from registry, and chains by instantiation

drmaas:
This is useful in 2 cases:
- using chains to filter all requests
- installing handlers and chains from external modules

---

This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1041)


drmaas:
Build appears to have failed on an unrelated module.

alkemist:
Looks good. Please add `@since 1.5` to the new methods in the Javadoc.

PR #1047, opened 2016-08-17
Add HttpClientReadTimeoutException

### LINKED ISSUE #1046: ReadTimeoutException results in raw RuntimeException, it should be subclassed
Currently, when a netty `ReadTimeoutException` is thrown, it [throws a raw RuntimeException](https://github.com/ratpack/ratpack/blob/62d4be299ac55397f3a6d8d781bbe5af22fab40f/ratpack-core/src/main/java/ratpack/http/client/internal/RequestActionSupport.java#L178).

This should be a subclass of `RuntimeException` that end users can look for (in my case, so that I can retry the request with a time delay).

As it is now, I need to look for a `RuntimeException` with a regex looking at the error message to determine if that was the type of exception.

beckje01:
HttpClientReadTimeoutException replaces the RuntimeException in the case
of a read timeout in the HttpClient.

This fixes #1046

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1047)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/client/HttpClientReadTimeoutException.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/HttpClientReadTimeoutException.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


alkemist:
@beckje01 can you please rebase the changes on to the rel_1.4.1 branch.

beckje01:
Snap can't seem to see that the target branch has changed. reopening

PR #1048, opened 2016-08-17
Upgrade to pac4j v1.9.4

leleuj:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1048)


ratpack-pac4j/src/main/java/ratpack/pac4j/RatpackPac4j.java
```diff
@@ -201,17 +202,28 @@ public static Handler authenticator(String path, ClientsProvider clientsProvider
    */
   @SafeVarargs
   @SuppressWarnings("varargs")
-  public static <C extends Credentials, U extends UserProfile> Handler requireAuth(Class<? extends Client<C, U>> clientType, Authorizer<? super U>... authorizers) {
+  public static <C extends Credentials, U extends CommonProfile> Handler requireAuth(Class<? extends Client<C, U>> clientType, Authorizer<? super U>... authorizers) {
```
danveloper: This looks like a binary incompatible change

danveloper:
Thanks for the PR @leleuj! It looks like there are breaking changes here. Is it possible to preserve the use of `UserProfile` (`CommonProfile` seems to extend `UserProfile`) or are you leveraging features of the new class in this change set?

leleuj:
In fact, the bound of the `Client` interface has changed from `UserProfile` to `CommonProfile`,  that's why I replaced it everywhere. I think it's better than having both, but I can limit the replacement to the places where it's really needed.

I just saw I need to fix the build also.

leleuj:
I think I've fixed the build, but it times out:

```
No console output received from command for the last 15 minutes.
Check out http://docs.snap-ci.com/troubleshooting/#my-build-is-timing-out for more info.
```

yterradas:
is there any updates on this issue?

leleuj:
I just updated the links to the new wiki. For me, everything is ok, the slightly breaking change (`UserProfile` -> `CommonProfile`) is necessary since the `Client` method signature has changed.

Ready for merge

leleuj:
@danveloper @alkemist Can we proceed with the merge? The build fails on timeout but I'm not sure if / how it's related.

danveloper:
The breaking change is a show stopper at the moment. We will have to figure out a way to address backward compatibility with existing applications.

leleuj:
OK. I understand. But this is why major versions exist, to say things have changed. It should be merged in a major Ratpack upgrade then.

Practically, `CommonProfile` inherits from `UserProfile` so I doubt there will be major impacts for most users (except compilation failures).

leleuj:
BTW, I just upgraded to the new 1.9.2 release: http://www.pac4j.org/docs/release-notes.html

yterradas:
@danveloper what is the plan for merging this issue? pac4j 1.9.x makes it a lot easier to provide custom encoders because `matches` is part of `PasswordEncoder` interface.  it also comes with BCrypt encoder.

danveloper:
@leleuj is it possible we can changed the signatures that changed from `UserProfile` => `CommonProfile` back to `UserProfile` and just do instancetype checks for the areas we specifically need it. The concern is strictly with binary compatibility and backwards compatibility. I'm happy to help make those changes in this PR if that's a viable solution?

leleuj:
I understand your concern.

Regarding the move from `UserProfile` to `CommonProfile`, I don't think it will ever generate any breaking change because the `UserProfile` class still exists (even if it's now abstract) and no one uses the `Client` interface or `BaseClient` abstract class directly.
But indeed, there are breaking changes: some components now throw a `HttpAction` while they did not, `RequiresHttpAction` was renamed as `HttpAction`...: there would be so much more to revert.

I do think this is acceptable, but minor adjustments may be required.

We cannot freeze APIs for ever: we try to limit breaking changes, but major versions can have breaking changes by definition. And I'm sure Ratpack has the same concern. So we must be able to handle these use cases.

Cannot we upgrade to a major pac4j version only for a major Ratpack version (major meaning with breaking changes)?

danveloper:
> I don't think it will ever generate any breaking change because the UserProfile class still exists...

Right, it won't be binary incompatible, but it still is a breaking change. User code will need to be updated between minor releases, and that's not desirable.

>... RequiresHttpAction was renamed as HttpAction...

This is a binary incompatible change between minor versions of Pac4j, which we can't inherit into Ratpack.

>We cannot freeze APIs for ever: we try to limit breaking changes, but major versions can have breaking changes by definition. And I'm sure Ratpack has the same concern. So we must be able to handle these use cases.

Totally agree, but we can't break APIs between minor versions. Users need a guarantee that their code will work and be compatible in the 1.x.x line.

We're currently on Pac4j 1.8.8, and 1.9.2 is a minor version update that introduces both binary incompatible and breaking changes in the API. For the current state of things, we're stuck on 1.8.8.

>Cannot we upgrade to a major pac4j version only for a major Ratpack version (major meaning with breaking changes)?

Absolutely, but we're pretty far off from Ratpack 2.x today, and unfortunately these breaking changes from Pac4j are manifesting in minor version releases, not major releases. If the 1.x.x release of Pac4j could provide a guarantee to binary and API compatibility, then we could easily transition Ratpack through the minor updates.

It may be too late for this, but what I would have suggested was creating a `1_X_X` branch in Pac4j where you could provide non-breaking updates and bug fixes to anything 1.x.x, and when you felt the API and class structures needed to be changed, you could update to 2.x.x and support all minor releases in that track.

As a user of Ratpack and Pac4j, I can say that the breaking changes from 1.7.x => 1.8.x were painful to work with in user code. Luckily, we didn't have to break any Ratpack APIs to bring 1.8.x in, and in this case I'm not super eager to put that tax on our users.

@leleuj, what I think would be a better approach to getting these changes in would be to have Pac4j set its major version to 2.0.0 and we create a `ratpack-pac4j-2` module for the new version line. We would need to be sure though that going forward the 2.x.x line wouldn't break compatibility between minor version. The existing `ratpack-pac4j` module would thus be stuck at 1.8.8, where we are today, but we could deprecate the existing stuff and direct users to move to the `ratpack-pac4j-2` module.

What are your thoughts on this?

leleuj:
>
> We're currently on Pac4j 1.8.8, and 1.9.2 is a minor version update that
> introduces both binary incompatible and breaking changes in the API. For
> the current state of things, we're stuck on 1.8.8.
>
I think there is a misunderstanding with the pac4j versioning: 1.0, 1.1,
1.2...1.8 and 1.9 are major versions with breaking changes. Using semver,
we would talk about versions 8.8 and 9.2. I'm seriously thinking about
changing version numbers to make that clear.


> Cannot we upgrade to a major pac4j version only for a major Ratpack
> version (major meaning with breaking changes)?
>
> Absolutely, but we're pretty far off from Ratpack 2.x today, and
> unfortunately these breaking changes from Pac4j are manifesting in minor
> version releases, not major releases. If the 1.x.x release of Pac4j could
> provide a guarantee to binary and API compatibility, then we could easily
> transition Ratpack through the minor updates.
>
This is guaranteed for 1.9.0, 1.9.1, 1.9.2 but not with 1.8.x due to our
versioning.


> It may be too late for this, but what I would have suggested was creating
> a 1_X_X branch in Pac4j where you could provide non-breaking updates and
> bug fixes to anything 1.x.x, and when you felt the API and class structures
> needed to be changed, you could update to 2.x.x and support all minor
> releases in that track.
>
This is an option: we could backport some of the evolutions of 1.9.x to
1.8.x, I have received some inquiries to release a 1.8.10 and it's
definitely not a problem.


> As a user of Ratpack and Pac4j, I can say that the breaking changes from
> 1.7.x => 1.8.x were painful to work with in user code. Luckily, we didn't
> have to break any Ratpack APIs to bring 1.8.x in, and in this case I'm not
> super eager to put that tax on our users.
>
I'm surprised of the "painful" aspect of the upgrades: I handle most of the
demo upgrades from version 1.0 and it's nothing more package name updates,
classes renaming...


> @leleuj <https://github.com/leleuj>, what I think would be a better
> approach to getting these changes in would be to have Pac4j set its major
> version to 2.0.0 and we create a ratpack-pac4j-2 module for the new
> version line. We would need to be sure though that going forward the 2.x.x
> line wouldn't break compatibility between minor version. The existing
> ratpack-pac4j module would thus be stuck at 1.8.8, where we are today,
> but we could deprecate the existing stuff and direct users to move to the
> ratpack-pac4j-2 module.
>
This is an option. Though, the same issue will arise with pac4j v2.0 and
v2.1 and v2.2...
So either we backport more and Ratpack 1.x stays on pac4j 1.8.x and then
Ratpack 2.x will be based on pac4j v1.9 or v2.0 or v2.1 (9.x, 10.x or 11.x
using semver) or we extract the ratpack-pac4j module as a separate library
to let it have its own versioning and we'll be soon at version 5 or 6 while
Ratpack will only reach version 2. I'm reluctant to have ratpack-pac4j as a
separate library as I prefer the Rapack team to maintan it, but if you're
willing to work like it, I can deal with it.

davidmc24:
If we have a module-per-pac4j major version, then this won't be a problem for future versions, other than the need to create the modules.  However, I think that for the foreseeable future, pac4j might be far more likely to release new major versions than ratpack.  We might want to consider deprecating the ratpack-pac4j module shipped with ratpack, and instead work towards supporting new versions via a pac4j-ratpack-1 module published by pac4j.

leleuj:
I'm in line with a separate module for ratpack-pac4j to accommodate the different lifecycles. I'd prefer it to be hosted on the Ratpack organization as it's officially supported by Ratpack. I'll continue to help for upgrades of course, but you'll remain in charge of the releases.

What do you think?

johnrengelman:
I also agree that there should be a separate `ratpack-pac4j` module per `pac4j` major version (so we should have done a 1.8 and a 1.9). I think the confusion here stems from the fact that SEMVER says that 1.8 -> 1.9 is a minor version not a major and thus shouldn't have breaking changes (http://semver.org/).

Ultimately, if a pac4j release requires any modification of a user's ratpack code base other than updating the dependency and recompiling, then it should be released as a new ratpack module.

As for where that lives, based on the release cadence of pac4j w.r.t. to breaking changes, I don't see it being feasible for it to continue living in the ratpack main module if we go this route. We have to remember that anytime we modifying internal code of ratpack, we have to ensure that the modules keep working, so as we add more versions of pac4j, that's more and more modules to validate/update to work with new the new internal coding.

I'd prefer to pick a major version of pac4j and say that that is the version that is going to be supported in ratpack until ratpack 2. If there is a desire for support for new versions, then that would need to come from an external module.

My 2 cents.

leleuj:
If we all agree, I let you set up a new `ratpack-pac4j` library under the `ratpack` organization...

KangoV:
Will there be any movement on this as 1.9.1 fixes this horrible little bug in 1.8.8 : https://github.com/pac4j/pac4j/issues/500

leleuj:
I just setup: https://github.com/pac4j/ratpack-pac4j

If someone is willing to port the current source code of the ratpack-pac4j module to this new project, we'll then upgrade it to pac4j v1.9.8 for the v1.5.0 release. Finally, we'll create and release the v2.0.0 based on pac4j v2.0.0.

Issue #1021, opened 2016-07-18
Update to Rx 1.1.7 and integrate Single and other new types


alkemist:

alkemist:
@tednaleid interested in looking at this one?

tednaleid:
@alkemist yep, I can work on this

tednaleid:
As part of this work, we'll want to update the `RxRatpack.initialize()` method so that it works properly with scripts that use the `@Grab` annotation to run ratpack.

In current scripts that use RxJava 1.1.2, using the `@Grab` annotation causes the `rx.Observable` class to get loaded in the classloader.  This causes the [static initializers to run and populate the `hook` field](https://github.com/ReactiveX/RxJava/blob/v1.1.2/src/main/java/rx/Observable.java#L63).  That happens before we have any chance to run `RxRatpack.initialize()` and registers the "wrong" execution hooks so groovy scripts are not fully ratpack aware.

This example script shows the issue using groovy 2.4.7:

```
@Grab('io.ratpack:ratpack-groovy:1.3.3')
@Grab('io.ratpack:ratpack-rx:1.3.3')

import ratpack.rx.RxRatpack
import static ratpack.groovy.Groovy.ratpack

RxRatpack.initialize()

ratpack {
  handlers {
    get { render "Hello World" }
  }
}
```

throws:

```
java.lang.IllegalStateException: Cannot install RxJava integration because another execution hook (class rx.plugins.RxJavaObservableExecutionHookDefault) is already installed
	at ratpack.rx.RxRatpack.initialize(RxRatpack.java:102)
	at ratpack.rx.RxRatpack$initialize.call(Unknown Source)
	at ConsoleScript0.run(ConsoleScript0:7)
```

So it has `RxJavaObservableExecutionHookDefault` installed (and unable to be overridden, instead of `rx.plugins.RxJavaSchedulersHook` that we want.

There are no public methods that enable the removal/resetting of this value, but in RxJava 1.1.7, this plugin system has been refactored to [delegate hooks](https://github.com/ReactiveX/RxJava/blob/v1.1.8/src/main/java/rx/Observable.java#L98) to a new [`RxJavaHooks`](https://github.com/ReactiveX/RxJava/blob/v1.1.8/src/main/java/rx/plugins/RxJavaHooks.java#L246) class that allows the app to `clear`, `reset`, and `lockdown` the hooks.

[...]

PR #1062, opened 2016-09-11
upgrading to RxJava 1.1.10

tednaleid:
❗ This PR that is meant to address #1021 isn't finished and shouldn't be merged yet.

I wanted to start a discussion about a few things that I've run into while upgrading RxJava and what others think is the right way to move forward.

The current blocker that I want to discuss is this error during the `gradle check`:

```
ratpack.manual.JavadocCodeSnippetTests > ratpack.rx.RxRatpack:448 FAILED
    ratpack.manual.snippets.executer.CompileException: java.lang.AssertionError: MANDATORY_WARNING: <R>extend(rx.functions.Func1<? super rx.Observable.OnSubscribe<T>,? extends R>) in rx.Observable has been deprecated (10:57)

        Caused by:
        java.lang.AssertionError: MANDATORY_WARNING: <R>extend(rx.functions.Func1<? super rx.Observable.OnSubscribe<T>,? extends R>) in rx.Observable has been deprecated (10:57)
```

(this error happens in a few other places)

The methods that this occurs on have javadoc examples using the `rx.Observable.extends` method which was previously `@Experimental` and [is now `@Deprecated`](https://github.com/ReactiveX/RxJava/blob/v1.1.10/src/main/java/rx/Observable.java#L215).  That is causing the javadoc examples to throw an error.  

The methods that these are related to are all versions of methods that are "intended to be used in conjunction with the `Observable#extend` method as a method reference".  That makes me think that we should deprecate our version of these methods as well and then mark these particular failures as ones to ignore.

I've tentatively marked them as `@Deprecated` in this PR but was curious what others thought.  I don't use them in my code base, but there could be a use that I'm not seeing.

---

This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1062)


ratpack-rx/src/main/java/ratpack/rx/RxRatpack.java
```diff
@@ -346,6 +330,7 @@ public static void initialize() {
    * @see #promise(Observable)
    * @throws UnmanagedThreadException if called outside of an execution
    */
+  @Deprecated
```
tednaleid: should we deprecate these methods?  the `rx.Observable.extend` method is deprecated and marked for removal in RxJava 1.2, but that appears to be the primary use case for these.

johnrengelman:
Closing this as we now have the `ratpack-rx2` module which supports RxJava 2.x

Issue #1079, opened 2016-10-12
Invalid percent-encoding sequences in paths when binding to path tokens throw IllegalArgumentException


pocockn:
Sending requests to paths like `networking%party` which obviously contain bad percent-encoding sequences is a client error and should most probably result in a 400 response. Currently `IllegalArgumentException` is thrown from within `TokenPathBiding.decodeURIComponent(String)` which is very hard to be handled by the application if it would like to send a 400 response in such situation.

Possible solutions:
- Ratpack should respond with 400 in such situations without any action necessary from the application developer
- Wrap the `IllegalArgumentException` in a more specific exception that can then be tested for in `ServerErrorHandler` implementation and handled appropriately

A failing test case to follow.

danveloper:
I'm in favor of the 2nd option. Would you be willing to put together a PR for this?

alkemist:
Dan, what's the basis of your preference? 

I can see that it's moe extensible in that user's could deal with it how they wish. On the flipside, hardly anyone is going to think to or remember to put in specific handling for this exception type and issue a 400.

I think the solution has to involve Ratpack issuing a 400, unless the app has specifically declared it wants to intervene here.

erdi:
How would the app declare that it wants to intervene? Would this be implemented by introducing a new interface similar to `ServerErrorHandler` with a default impl raising a 400 client error and users being able to provide their impl that does something different?

alkemist:
Probably something like that, yes. That would probably be best, and the default can just `ctx.clientError(400)`.

danveloper:
@alkemist I'm not hard-lined on my thoughts on this, but my basis is that a `400` seems to me like a response code that should come from user code. If we wrap the exception, then apps can easily translate it into a `400` on their side. But since it's technically not a valid request in the first place, I can see it both ways... I don't really know.

alkemist:
> but my basis is that a 400 seems to me like a response code that should come from user code

I get that spirit, but I don't think it will work in practice. Most people aren't going to think of this until it bites them.

The default behaviour should be that there is an indication to the client that they did something wrong. The only way to do this would be to issue a 4xx.

We need to see what other frameworks and HTTP servers do when they receive malformed request paths.

PR #1080, opened 2016-10-12
Failing test case for ratpack/ratpack#1079

pocockn:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1080)

### DIFF
--- a/ratpack-core/src/test/groovy/ratpack/path/PathBindingSpecSpec.groovy
+++ b/ratpack-core/src/test/groovy/ratpack/path/PathBindingSpecSpec.groovy
@@ -16,6 +16,7 @@
 
 package ratpack.path
 
+import ratpack.test.ApplicationUnderTest
 import ratpack.test.internal.RatpackGroovyDslSpec
 
 class PathBindingSpecSpec extends RatpackGroovyDslSpec {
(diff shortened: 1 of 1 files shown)


ratpack-core/src/test/groovy/ratpack/path/PathBindingSpecSpec.groovy
```diff
@@ -53,4 +54,31 @@ class PathBindingSpecSpec extends RatpackGroovyDslSpec {
...
+  def "binding incorrectly formatted paths to tokens should not cause server errors"() {
+    when:
+    handlers {
+      get(":token") {
+        response.send()
+      }
+    }
+
+    then:
+    statusCode(applicationUnderTest, "networking%party") != 500
```
danveloper: you can just do `get("networking%party").statusCode` here

erdi: No, you can't. If you could we would do that. :)

The problem is that `TestHttpClient` will not let you send request for illegal paths by throwing an exception, but there is nothing in the http protocol to prevent you from sending illegal paths which is what we are doing in `statusCode()`.

Maybe this justifies getting `TestHttpClient` to allow sending bad paths like these using some new method?

PR #1083, opened 2016-10-18
Include URL query string in NCSA request log

mkavanagh:
For our use case the query string is often very important in understanding what requests were made but it's not currently included by the NCSA-style request logger (which includes only the path portion of the request URI). My understanding is that by convention the URL logged in NCSA-style access logs is the literal URL from the first line of the HTTP request (including any query string), so that's what I've changed it to do.

---

This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1083)


alkemist:
Thanks @mkavanagh .

Sorry it took so long.

Issue #1084, opened 2016-10-19
Add Promise.nextOpIf(Predicate<? super T> predicate, Function<? super T,? extends Operation> function)


brucenunk:
This would be useful to conditionally execute an Operation. It can be implemented currently using the following but it is very noisy.

```
promise
.flatMapIf({ condition }) { value ->
    operation.map { value }
}
```

alkemist:
Sounds good. Are you up for adding it?

brucenunk:
Yeah sure. I implemented it locally last night but will create a pull request today.

brucenunk:
@alkemist I've submitted a PR for this, https://github.com/ratpack/ratpack/pull/1086

Broke the javadoc tests initially, then screwed up the JavaDoc by running "reformat code" in IntelliJ but it's turned out OK in the end ... I think :)

PR #1086, opened 2016-10-20
Add Promise.nextOpIf(Predicate<? super T> predicate, Function<? super T,? extends Operation> function)

brucenunk:
Implemented Promise#nextOpIf as per https://github.com/ratpack/ratpack/issues/1084

---

This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1086)


johnrengelman:
This is going to be handled in `1.5` with the addition of `Function.conditional` and `Action.conditional`

See https://github.com/ratpack/ratpack/commit/725247055bac6808833cfe522419b664fc102e69 and https://github.com/ratpack/ratpack/commit/cc337c34b28927c7dd7fe7036aabca2f0cb260e4

johnrengelman:
@ratpack/committers - do we want to keep adding syntatic sugar like this after introduction the above `conditional` methods on `Function` and `Action`?

rhart:
What I like about the sugar is that it makes discovering these capabilities easy, it's explicitly part of the api.  

Even adding to the documentation and providing more examples, the capabilities would become more hidden IMO.

PR #1097, opened 2016-10-31
Introduce InvalidPathEncodingErrorHandler

pocockn:
InvalidPathEncodingErrorHandler allows to deal with requests to incorrectly encoded paths. The default implementation responds with a 400 status code.

This PR provides a fix for #1079.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1097)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/error/InvalidPathEncodingErrorHandler.java
+++ b/ratpack-core/src/main/java/ratpack/error/InvalidPathEncodingErrorHandler.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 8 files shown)


ratpack-core/src/main/java/ratpack/error/InvalidPathEncodingErrorHandler.java
```diff
@@ -0,0 +1,50 @@
...
+   * <p>
+   * Implementations should strive to avoid throwing exceptions.
+   * If exceptions are thrown, they will just be logged at a warning level and the response will be finalised with a 500 error code and empty body.
+   *
+   * @param context The context being processed
+   * @param exception The path encoding error that occurred
+   * @throws Exception if something goes wrong handling the error
+   */
+  @NonBlocking
+  void error(Context context, InvalidPathEncodingException exception) throws Exception;
```
erdi: @alkemist, one thing to consider is that instead of introducing a completely new handler type we could add a default method with this signature to `ServerErrorHandler` and use the implementation that is currently in `DefaultInvalidPathEncodingErrorHandler`.

alkemist:
Very nicely put together contribution. Thanks @pocockn.

I need another day to think about whether this is the way to handle this. At this stage, I'm inclined to merge go this route.

PR #1111, opened 2016-11-17
Signal subscription cancellation when channel is abruptly closed.

niwinz:
I have experimented in some situations when the publisher subscription is not properly closed. It happens as example when SSE connection is established but no data is transferred, then the connection is abruptly closed. In this case, the cancel is never called causing subscriptions leaking.

This PR solves it adding an additional listener to closeFuture.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1111)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/server/internal/DefaultResponseTransmitter.java
+++ b/ratpack-core/src/main/java/ratpack/server/internal/DefaultResponseTransmitter.java
@@ -171,6 +171,10 @@ public class DefaultResponseTransmitter implements ResponseTransmitter {
         }
       };
 
+      private final ChannelFutureListener onCloseListener = future -> {
+        cancel();
+      };
+
       private void cancel() {
         if (done.compareAndSet(false, true)) {
           subscription.cancel();
@@ -202,6 +206,7 @@ public class DefaultResponseTransmitter implements ResponseTransmitter {
           notifyListeners(responseStatus, channel.close());
         } else {
           channelFuture.addListener(cancelOnFailure);
+          channel.closeFuture().addListener(onCloseListener);
           if (channel.isWritable()) {
             this.subscription.request(1);
           }


niwinz:
Seems like the tests are failing for unrelated reasons. 

Any hope to get this reviewed, this is very annoying and forces restart the application periodically because of connection leaks.

danveloper:
could this perhaps come with a test demonstrating the problem?

niwinz:
I'm not very fluent with ratpack test suite :(, but I can provide a reproducible test application and the steps.

Clone https://github.com/niwinz/ratpack-playground repository (in master branch). It contains a small application that sends a never finishing stream (just for emulate the situation when the sse connection is opened and no data is sent).

Run the app in one terminal:

```bash
cd ratpack-payload
./gradlew --no-daemon clean run
```

Then in other command open a curl and interrupt it after some seconds:

```bash
curl -v http://localhost:5050/
```

Without the patch applied, yo will see the following output on the terminal where the server is running:

```text
[main] INFO ratpack.server.RatpackServer - Ratpack started (development) for http://localhost:5050

TEST: request(1)
```

You will observe that no cancel is called. If you apply the patch, the output will be the expected:

```text
[main] INFO ratpack.server.RatpackServer - Ratpack started (development) for http://localhost:5050

TEST: request(1)
TEST: cancel
```

It only happens when no first message is sent...

niwinz:
@danveloper bump

alkemist:
@niwinz sorry, I forgot this existed. I fixed this as #1150.

niwinz:
Thanks! ETA for the next release? I'm waiting this fix for a while (it is currently blocking release of catacumba).

alkemist:
it's actually out now.

niwinz:
many thanks!, it is published on maven?

PR #1115, opened 2016-11-23
Change dependency example in gradle section

### LINKED ISSUE #1101: ratpack.dependency("jackson") requires non-existent dependency
When I put this into build.gradle:

dependency {
  compile ratpack.dependency("jackson")
}

I get an error message saying that ratpack-jackson:1.3.3 is required. However, there is no such version in the repositories. The latest version in MVN is ratpack-jackson:0.9.17.

beckje01:
Removed refrences to the old ratpack.dependency("jackson") and used the
metrics lib instead.

This fixes #1101

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1115)

### DIFF
--- a/ratpack-manual/src/content/chapters/60-gradle.md
+++ b/ratpack-manual/src/content/chapters/60-gradle.md
@@ -72,11 +72,11 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


danveloper:
👍

PR #1117, opened 2016-11-26
Surface error when request bodyStream is closed unexpectedly

### LINKED ISSUE #1112: Request.getBodyStream is never completed if client drops connection
If client drops connection unexpectedly, Request.getBodyStream is never completed (neither with error nor with success).
I'm not sure how to create unit test for network failures, so here is how to reproduce with real server:
1. Start server from this source: https://gist.github.com/jurna/d23c0ec3144a005dd44691a77a21522b
2. Post fairly big file
```
curl -X POST "http://localhost:5050/" --data-binary @someBigFile.tar.gz
```
3. Hit ctrl+c before curl finishes file upload.
4. Check the log files - no onComplete or onError is invoked.

danveloper:
This PR surfaces an error to a request bodyStream `Subscriber` when the client closes the connection before the content-length is reached.

* Fixes #1112

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1117)

### DIFF
--- a/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy
+++ b/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy
@@ -19,6 +19,7 @@ package ratpack.http
 import io.netty.buffer.ByteBuf
 import io.netty.buffer.ByteBufUtil
 import io.netty.buffer.Unpooled
+import io.netty.util.CharsetUtil
 import org.apache.commons.lang3.RandomStringUtils
 import org.reactivestreams.Subscriber
 import org.reactivestreams.Subscription
@@ -28,8 +29,10 @@ import ratpack.http.client.RequestSpec
 import ratpack.registry.Registry
 import ratpack.stream.Streams
 import ratpack.test.internal.RatpackGroovyDslSpec
+import spock.lang.Timeout
 
 import java.nio.charset.StandardCharsets
+import java.util.concurrent.CountDownLatch
 
 class RequestBodyStreamReadingSpec extends RatpackGroovyDslSpec {
 
@@ -343,6 +346,65 @@ class RequestBodyStreamReadingSpec extends RatpackGroovyDslSpec {
     requestSpec { it.body.text string }
     postText() == "ok"
     file.text == string
+  }
+
+  @Timeout(10)
+  def "request body stream errors when client closes connection unexpectedly"() {
+    when:
+    def subscribed = new CountDownLatch(1)
+    def error = new CountDownLatch(1)
+
+    handlers {
+      all { ctx ->
+        request.getBodyStream(Long.MAX_VALUE).subscribe(new Subscriber<ByteBuf>() {
+          Subscription subscription
+
+          @Override
+          void onSubscribe(Subscription s) {
+            subscribed.countDown()
+            subscription = s
+            subscription.request 1
+          }
+
+          @Override
+          void onNext(ByteBuf byteBuf) {
+            byteBuf.release()
+            subscription.request 1
+          }
+
+          @Override
+          void onError(Throwable t) {
+            error.countDown()
+            ctx.error(t)
+          }
+
+          @Override
+          void onComplete() {
+
+          }
+        })
+      }
+    }
+
+    and:
+    Socket socket = new Socket()
+    socket.connect(new InetSocketAddress(address.host, address.port))
+    new OutputStreamWriter(socket.outputStream, "UTF-8").with {
+      write("POST / HTTP/1.1\r\n")
+      write("Content-Type: multipart/form-data;boundary=---foo\r\n")
+      write("Content-Length: 4000\r\n")
+      write("\r\n")
+      close()
+    }
+
+    and:
+    socket.close()
+    subscribed.await()
+    error.await()
+
+    then:
+    subscribed.count == 0
+    error.count == 0
   }
 
   class BufferThenSendSubscriber implements Subscriber<ByteBuf> {
--- a/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy
+++ b/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy
@@ -19,7 +19,6 @@ package ratpack.http
 import io.netty.buffer.ByteBuf
 import io.netty.buffer.ByteBufUtil
 import io.netty.buffer.Unpooled
-import io.netty.util.CharsetUtil
 import org.apache.commons.lang3.RandomStringUtils
 import org.reactivestreams.Subscriber
 import org.reactivestreams.Subscription
--- a/ratpack-core/src/main/java/ratpack/http/ConnectionClosedException.java
+++ b/ratpack-core/src/main/java/ratpack/http/ConnectionClosedException.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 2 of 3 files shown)


ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy
```diff
@@ -19,6 +19,7 @@ package ratpack.http
 import io.netty.buffer.ByteBuf
 import io.netty.buffer.ByteBufUtil
 import io.netty.buffer.Unpooled
+import io.netty.util.CharsetUtil
```
johnrengelman: unused import

rhart:
*[ratpack-core/src/main/java/ratpack/server/internal/RequestBody.java, line 158 at r2](https://reviewable.io:443/reviews/ratpack/ratpack/1117#-KXzuKZru-TzDN_oNfQF:-KXzuKZru-TzDN_oNfQG:b2p7s04) ([raw file](https://github.com/ratpack/ratpack/blob/6f25a2710d1018879b8cb80153b85244fcd1648b/ratpack-core/src/main/java/ratpack/server/internal/RequestBody.java#L158)):*
> ```Java
>       return new Subscription() {
>         boolean autoRead;
>         boolean unexpectedCloseHandled;
> ```

If this is here to stop us adding the listener twice, should it be AtomicBoolean?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1117)*

rhart:
*[ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy, line 375 at r2](https://reviewable.io:443/reviews/ratpack/ratpack/1117#-KXzulA5DhzO9HV_3jNM:-KXzulA5DhzO9HV_3jNN:b-kq1hgu) ([raw file](https://github.com/ratpack/ratpack/blob/6f25a2710d1018879b8cb80153b85244fcd1648b/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy#L375)):*
> ```Groovy
> 
>           @Override
>           void onError(Throwable t) {
> ```

It would be good to assert that the error is what we expect.  Either as an `assert` here or assign a variable and check in the `then`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1117)*

rhart:
*[ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy, line 401 at r2](https://reviewable.io:443/reviews/ratpack/ratpack/1117#-KXzuzL0maQzRm7t_bzl:-KXzuzL0maQzRm7t_bzm:bir5dfg) ([raw file](https://github.com/ratpack/ratpack/blob/6f25a2710d1018879b8cb80153b85244fcd1648b/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy#L401)):*
> ```Groovy
>     and:
>     socket.close()
>     subscribed.await()
> ```

Why do we need to wait here?  Did you mean this to be before `socket.close()` is called

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1117)*

rhart:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1117)*

alkemist:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[ratpack-core/src/main/java/ratpack/server/internal/RequestBody.java, line 158 at r2](https://reviewable.io:443/reviews/ratpack/ratpack/1117#-KXzuKZru-TzDN_oNfQF:-KfdpgecsvAfUemsNatX:beuuz9j) ([raw file](https://github.com/ratpack/ratpack/blob/6f25a2710d1018879b8cb80153b85244fcd1648b/ratpack-core/src/main/java/ratpack/server/internal/RequestBody.java#L158)):*
<details><summary><i>Previously, rhart (Rus Hart) wrote…</i></summary><blockquote>

If this is here to stop us adding the listener twice, should it be AtomicBoolean?
</blockquote></details>

This is going to move where it's not under contention. See below.

---

*[ratpack-core/src/main/java/ratpack/server/internal/RequestBody.java, line 162 at r2](https://reviewable.io:443/reviews/ratpack/ratpack/1117#-KfdpV966XyHY3MWrCJt:-KfdpV966XyHY3MWrCJu:bycffxr) ([raw file](https://github.com/ratpack/ratpack/blob/6f25a2710d1018879b8cb80153b85244fcd1648b/ratpack-core/src/main/java/ratpack/server/internal/RequestBody.java#L162)):*
> ```Java
>         @Override
>         public void request(long n) {
>           if (!unexpectedCloseHandled) {
> ```

We should move this up to just after we set autoRead = false. We are allowed to emit an error at any time.

---

*[ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy, line 354 at r2](https://reviewable.io:443/reviews/ratpack/ratpack/1117#-KfdqQ0gZXxm4zy7lsJn:-KfdqQ0gZXxm4zy7lsJo:bdqoz44) ([raw file](https://github.com/ratpack/ratpack/blob/6f25a2710d1018879b8cb80153b85244fcd1648b/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy#L354)):*
> ```Groovy
>     when:
>     def subscribed = new CountDownLatch(1)
>     def error = new CountDownLatch(1)
> ```

We should be able to do away with both of these latches.

---

*[ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy, line 375 at r2](https://reviewable.io:443/reviews/ratpack/ratpack/1117#-KXzulA5DhzO9HV_3jNM:-KfdqEI4pF_QX7R2ii_R:b-aaiu5p) ([raw file](https://github.com/ratpack/ratpack/blob/6f25a2710d1018879b8cb80153b85244fcd1648b/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy#L375)):*
<details><summary><i>Previously, rhart (Rus Hart) wrote…</i></summary><blockquote>

It would be good to assert that the error is what we expect.  Either as an `assert` here or assign a variable and check in the `then`
</blockquote></details>

Elsewhere, we just render the exception class name and then assert on that response.

---

*[ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy, line 382 at r2](https://reviewable.io:443/reviews/ratpack/ratpack/1117#-KfdpuIFPtMr7CJsLYNW:-KfdpuIFPtMr7CJsLYNX:b-loixk9) ([raw file](https://github.com/ratpack/ratpack/blob/6f25a2710d1018879b8cb80153b85244fcd1648b/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy#L382)):*
> ```Groovy
>           @Override
>           void onComplete() {
> ```

Should do something like `render "ok"` here in case the test fails.

---

*[ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy, line 395 at r2](https://reviewable.io:443/reviews/ratpack/ratpack/1117#-Kfdq6qSEmzj20FRSajZ:-Kfdq6qSEmzj20FRSaj_:b-axqttq) ([raw file](https://github.com/ratpack/ratpack/blob/6f25a2710d1018879b8cb80153b85244fcd1648b/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy#L395)):*
> ```Groovy
>       write("Content-Type: multipart/form-data;boundary=---foo\r\n")
>       write("Content-Length: 4000\r\n")
>       write("\r\n")
> ```

Would be good to _also_ test with sending _some_ body.

---

*[ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy, line 401 at r2](https://reviewable.io:443/reviews/ratpack/ratpack/1117#-KXzuzL0maQzRm7t_bzl:-Kfdq4xQ3EMIJn4wofeV:b-ro07q4) ([raw file](https://github.com/ratpack/ratpack/blob/6f25a2710d1018879b8cb80153b85244fcd1648b/ratpack-core/src/test/groovy/ratpack/http/RequestBodyStreamReadingSpec.groovy#L401)):*
<details><summary><i>Previously, rhart (Rus Hart) wrote…</i></summary><blockquote>

Why do we need to wait here?  Did you mean this to be before `socket.close()` is called
</blockquote></details>

It shouldn't be needed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1117)*

PR #1118, opened 2016-11-26
Apply Spring Boot registry after Guice registry is built.

### LINKED ISSUE #1099: Unable to register ServerErrorHandler bean with ratpack springBoot
Ratpack version: 1.4.3
Spring boot version: 1.3.5.RELEASE

Registered a bean of type ServerErrorHandler and using @EnableRatpack annotation as per documentation, I should just able to register beans of type ServerErrorHandler and Ratpack will automatically register the bean into the registry and use that as the server handler rather than default. Could someone advise if this is incorrect or indeed a bug? I can provide code examples if needed.

Regards,

Joel

danveloper:
* This change allows Spring component bindings to be resolved first in a Ratpack / Spring Boot app that uses @EnableRatpack
* Fixes #1099

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1118)


johnrengelman:
LGTM

dopeymac:
@danveloper Do you know when this change will be released?

PR #1119, opened 2016-11-26
Ability to provide listener for unhandled execution errors

### LINKED ISSUE #1037: Support for execution error handlers from user registry
We should support the ability to add one or more execution error handlers by way of the user registry.

danveloper:
* Fixes #1037 

This commit provides the means by which users can add one or many `ExecutionErrorListener` instances to an execution registry. Furthermore it extends the capabilities of `ExecStarter#onError` to allow multiple error handlers to be specified in a forked execution.

Users will be able to use this feature to provide a handler that will capture any uncaught exceptions in an execution, and do further processing in a new execution.

### DIFF
--- a/ratpack-core/src/main/java/ratpack/exec/internal/DefaultExecController.java
+++ b/ratpack-core/src/main/java/ratpack/exec/internal/DefaultExecController.java
@@ -17,6 +17,7 @@
 package ratpack.exec.internal;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import io.netty.channel.EventLoop;
 import io.netty.channel.EventLoopGroup;
 import io.netty.util.concurrent.DefaultThreadFactory;
@@ -28,6 +29,7 @@ import ratpack.func.Block;
 import ratpack.registry.RegistrySpec;
 import ratpack.util.internal.ChannelImplDetector;
 
+import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -35,8 +37,6 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import static ratpack.func.Action.noop;
 
 public class DefaultExecController implements ExecControllerInternal {
-
-  private static final Action<Throwable> LOG_UNCAUGHT = t -> DefaultExecution.LOGGER.error("Uncaught execution exception", t);
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ExecController.class);
 
@@ -146,7 +146,7 @@ public class DefaultExecController implements ExecControllerInternal {
   @Override
   public ExecStarter fork() {
     return new ExecStarter() {
-      private Action<? super Throwable> onError = LOG_UNCAUGHT;
+      private List<ExecutionErrorListener> errorListeners = Lists.newArrayList();
       private Action<? super Execution> onComplete = noop();
       private Action<? super Execution> onStart = noop();
       private Action<? super RegistrySpec> registry = noop();
@@ -160,7 +160,7 @@ public class DefaultExecController implements ExecControllerInternal {
 
       @Override
       public ExecStarter onError(Action<? super Throwable> onError) {
-        this.onError = onError;
+        this.errorListeners.add((e, t) -> onError.execute(t));
         return this;
       }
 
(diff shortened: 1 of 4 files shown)


ratpack-core/src/main/java/ratpack/exec/internal/DefaultExecution.java
```diff
@@ -441,16 +462,34 @@ void enqueue(Block segment) {
...
-        LOGGER.error("error handler " + onError + " threw error (this execution will terminate):", errorHandlerError);
-        execStream = TerminalExecStream.INSTANCE;
-      }
+
+      final Execution currentExecution = DefaultExecution.this;
+      List<ExecutionErrorListener> errorListeners =
+        Lists.newArrayList(registry.getAll(ExecutionErrorListener.class));
+
+      controller.fork()
+        .onComplete(e -> {
```
danveloper: this allows us to pause the originating execution while the `ExecutionErrorListener` instances are working; then once they have completed, this will properly close off the original execution.

This is necessary to preserve the functionality that error handlers are invoked prior to the execution's `onComplete` handler.

ratpack-core/src/main/java/ratpack/exec/internal/DefaultExecController.java
```diff
@@ -160,7 +160,7 @@ public ExecStarter eventLoop(EventLoop eventLoop) {
 
       @Override
       public ExecStarter onError(Action<? super Throwable> onError) {
-        this.onError = onError;
+        this.errorListeners.add((e, t) -> onError.execute(t));
```
danveloper: with this, multiple `onError` calls can be made and they will all be fired on an error. This may constitute a breaking behavior change.

johnrengelman:
failing specs:
```
ratpack.exec.ExecutionErrorHandlingSpec > error listeners are pulled from registry FAILED
    Condition not satisfied:

    result == "!"
    |      |
    ""     false
           1 difference (0% similarity)
           (-)
           (!)
        at ratpack.exec.ExecutionErrorHandlingSpec.error listeners are pulled from registry(ExecutionErrorHandlingSpec.groovy:77)

ratpack.exec.ExecutionErrorHandlingSpec > multiple error listeners will be invoked and in registry order FAILED
    Condition not satisfied:

    results == [2, 1]
    |       |
    |       false
    [2, 1] (java.util.ArrayList)
        at ratpack.exec.ExecutionErrorHandlingSpec.multiple error listeners will be invoked and in registry order(ExecutionErrorHandlingSpec.groovy:101)

ratpack.manual.JavadocCodeSnippetTests > ratpack.exec.ExecutionErrorListener:27 FAILED
    ratpack.manual.snippets.executer.CompileException: java.lang.AssertionError: ERROR: illegal start of expression (29:12)

        Caused by:
        java.lang.AssertionError: ERROR: illegal start of expression (29:12)
```

ldaley:
I have merged master into this. Let's get this into 1.6.

ratpack-core/src/main/java/ratpack/exec/internal/DefaultExecController.java
ldaley: @danveloper why do we need to have this multiple error handler behaviour?

ratpack-core/src/main/java/ratpack/exec/internal/DefaultExecution.java
ldaley: We should try to avoid this. We already had the behaviour of executing the error handlers before the on complete handler. What broke about that?

ldaley:
@danveloper I would like to make a couple of changes.

1. Revert the additive behaviour of ExecSpec.onError
2. Avoid forking a new execution to invoke the error handlers

Other than that, I think we are good to merge (once we fix the tests).

danveloper:
For 1: how would you like to handle the addition of the error handlers? Via the registry?

For 2: how do we ensure the error handlers run inside an execution? The current execution is complete by the time the error handlers would be invoked.

ldaley:
1) I think we should restrict the goal here to just being the ability to override the “no error handler” error handler. So these would work just like interceptors/initializers in that we pull all implementations from the server registry. Add an impl to an execution while running or when bootstrapping would have no effect.

2) It should be possible to do this without forking a new execution.

I'll make both changes today.

gschrader:
Is this still queued up for 1.6?

beckje01:
@gschrader it's set to be reviewed this week and included if seems reasonable

beckje01:
I'll look at removing the additive behavior next then I think this will have all its tests passing.

michaelschlies:
@beckje01 is there outstanding things on this PR? running locally it isn't failing, and I've been seeing some strange timing issues out of circleci lately so it might be worth just re-triggering this to see if you just hit a hiccup.

PR #1124, opened 2016-11-28
Support for HTTPClient interceptors.

johnrengelman:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1124)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
@@ -75,7 +75,7 @@ public class DefaultHttpClient implements HttpClientInternal {
         .group(key.execution.getEventLoop())
         .channel(ChannelImplDetector.getSocketChannelImpl())
         .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, (int) key.connectTimeout.toMillis())
-        .option(ChannelOption.ALLOCATOR, byteBufAllocator)
+        .option(ChannelOption.ALLOCATOR, spec.byteBufAllocator)
         .option(ChannelOption.AUTO_READ, false)
         .option(ChannelOption.SO_KEEPALIVE, isPooling());
 
@@ -92,31 +92,15 @@ public class DefaultHttpClient implements HttpClientInternal {
     }
   };
 
-  private final ByteBufAllocator byteBufAllocator;
-  private final int maxContentLength;
-  private final int maxResponseChunkSize;
-  private final int poolSize;
-  private final Duration readTimeout;
-  private final Duration connectTimeout;
-  private final Action<? super RequestSpec> requestInterceptor;
-  private final Action<? super ReceivedResponse> responseInterceptor;
+  private final DefaultHttpClient.Spec spec;
 
-  private DefaultHttpClient(ByteBufAllocator byteBufAllocator, int maxContentLength, int maxResponseChunkSize, int poolSize, Duration readTimeout, Duration connectTimeout,
-                            Action<? super RequestSpec> requestInterceptor, Action<? super ReceivedResponse> responseInterceptor) {
-
-    this.byteBufAllocator = byteBufAllocator;
-    this.maxContentLength = maxContentLength;
-    this.maxResponseChunkSize = maxResponseChunkSize;
-    this.poolSize = poolSize;
-    this.readTimeout = readTimeout;
-    this.connectTimeout = connectTimeout;
-    this.requestInterceptor = requestInterceptor;
-    this.responseInterceptor = responseInterceptor;
+  private DefaultHttpClient(DefaultHttpClient.Spec spec) {
+    this.spec = spec;
   }
 
   @Override
   public int getPoolSize() {
-    return poolSize;
+    return spec.poolSize;
   }
 
   private boolean isPooling() {
@@ -130,33 +114,33 @@ public class DefaultHttpClient implements HttpClientInternal {
 
   @Override
   public Action<? super RequestSpec> getRequestInterceptor() {
-    return requestInterceptor;
+    return spec.requestInterceptor;
   }
 
   @Override
-  public Action<? super ReceivedResponse> getResponseInterceptor() {
-    return responseInterceptor;
+  public Action<? super HttpResponse> getResponseInterceptor() {
+    return spec.responseInterceptor;
   }
 
   public ByteBufAllocator getByteBufAllocator() {
-    return byteBufAllocator;
+    return spec.byteBufAllocator;
   }
 
   public int getMaxContentLength() {
-    return maxContentLength;
+    return spec.maxContentLength;
   }
 
   @Override
   public int getMaxResponseChunkSize() {
-    return maxResponseChunkSize;
+    return spec.responseMaxChunkSize;
   }
 
   public Duration getReadTimeout() {
-    return readTimeout;
+    return spec.readTimeout;
   }
 
   public Duration getConnectTimeout() {
-    return connectTimeout;
+    return spec.connectTimeout;
   }
 
   @Override
@@ -166,31 +150,19 @@ public class DefaultHttpClient implements HttpClientInternal {
 
   @Override
   public HttpClient copyWith(Action<? super HttpClientSpec> action) throws Exception {
-    DefaultHttpClient.Spec spec = new DefaultHttpClient.Spec();
-    Action<? super HttpClientSpec> clonedConfig = s -> {
-      s.byteBufAllocator(getByteBufAllocator());
-      s.maxContentLength(getMaxContentLength());
-      s.poolSize(getPoolSize());
-      s.readTimeout(getReadTimeout());
-      s.requestIntercept(getRequestInterceptor());
-      s.responseIntercept(getResponseInterceptor());
-    };
-    return of(clonedConfig.append(action));
+    return of(new DefaultHttpClient.Spec(spec), action);
   }
 
   public static HttpClient of(Action<? super HttpClientSpec> action) throws Exception {
     DefaultHttpClient.Spec spec = new DefaultHttpClient.Spec();
+    return of(spec, action);
+  }
+
+  private static HttpClient of(DefaultHttpClient.Spec spec, Action<? super HttpClientSpec> action) throws Exception {
     action.execute(spec);
 
     return new DefaultHttpClient(
-      spec.byteBufAllocator,
-      spec.maxContentLength,
-      spec.responseMaxChunkSize,
-      spec.poolSize,
-      spec.readTimeout,
-      spec.connectTimeout,
-      spec.requestInterceptor,
-      spec.responseInterceptor
+      spec
     );
   }
 
@@ -203,9 +175,20 @@ public class DefaultHttpClient implements HttpClientInternal {
     private Duration readTimeout = Duration.ofSeconds(30);
     private Duration connectTimeout = Duration.ofSeconds(30);
     private Action<? super RequestSpec> requestInterceptor = Action.noop();
-    private Action<? super ReceivedResponse> responseInterceptor = Action.noop();
+    private Action<? super HttpResponse> responseInterceptor = Action.noop();
 
     private Spec() {
+    }
+
+    private Spec(Spec spec) {
+      this.byteBufAllocator = spec.byteBufAllocator;
+      this.poolSize = spec.poolSize;
+      this.maxContentLength = spec.maxContentLength;
+	    this.responseMaxChunkSize = spec.responseMaxChunkSize;
+      this.readTimeout = spec.readTimeout;
+	    this.connectTimeout = spec.connectTimeout;
+      this.requestInterceptor = spec.requestInterceptor;
+      this.responseInterceptor = spec.responseInterceptor;
     }
 
     @Override
@@ -251,7 +234,7 @@ public class DefaultHttpClient implements HttpClientInternal {
     }
 
     @Override
-    public HttpClientSpec responseIntercept(Action<? super ReceivedResponse> interceptor) {
+    public HttpClientSpec responseIntercept(Action<? super HttpResponse> interceptor) {
       responseInterceptor = responseInterceptor.append(interceptor);
       return this;
     }
@@ -269,12 +252,12 @@ public class DefaultHttpClient implements HttpClientInternal {
 
   @Override
   public Promise<ReceivedResponse> request(URI uri, final Action<? super RequestSpec> requestConfigurer) {
-    return Promise.async(downstream -> new ContentAggregatingRequestAction(uri, this, 0, Execution.current(), requestConfigurer.append(requestInterceptor)).connect(downstream));
+    return Promise.async(downstream -> new ContentAggregatingRequestAction(uri, this, 0, Execution.current(), requestConfigurer.append(spec.requestInterceptor)).connect(downstream));
   }
 
   @Override
   public Promise<StreamedResponse> requestStream(URI uri, Action<? super RequestSpec> requestConfigurer) {
-    return Promise.async(downstream -> new ContentStreamingRequestAction(uri, this, 0, Execution.current(), requestConfigurer.append(requestInterceptor)).connect(downstream));
+    return Promise.async(downstream -> new ContentStreamingRequestAction(uri, this, 0, Execution.current(), requestConfigurer.append(spec.requestInterceptor)).connect(downstream));
   }
 
 }
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
@@ -186,9 +186,9 @@ public class DefaultHttpClient implements HttpClientInternal {
       this.byteBufAllocator = spec.byteBufAllocator;
       this.poolSize = spec.poolSize;
       this.maxContentLength = spec.maxContentLength;
-	    this.responseMaxChunkSize = spec.responseMaxChunkSize;
+      this.responseMaxChunkSize = spec.responseMaxChunkSize;
       this.readTimeout = spec.readTimeout;
-	    this.connectTimeout = spec.connectTimeout;
+      this.connectTimeout = spec.connectTimeout;
       this.requestInterceptor = spec.requestInterceptor;
       this.responseInterceptor = spec.responseInterceptor;
     }
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
@@ -22,6 +22,7 @@ import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.pool.*;
+import ratpack.exec.ExecController;
 import ratpack.exec.Execution;
 import ratpack.exec.Promise;
 import ratpack.exec.internal.ExecControllerInternal;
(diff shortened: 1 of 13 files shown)


rhart:
*[ratpack-core/src/main/java/ratpack/http/client/HttpClientSpec.java, line 82 at r1](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KXkqqY4kZVcj9sc1ilA:-KXkqqY4kZVcj9sc1ilB:b-qmd8ed) ([raw file](https://github.com/ratpack/ratpack/blob/675674c1beb888179002fc5ced0c1f7ce86c9289/ratpack-core/src/main/java/ratpack/http/client/HttpClientSpec.java#L82)):*
> ```Java
>    * Add an interceptor for all requests handled by this client.
>    * <p>
>    * This function is additive.
> ```

I read this to mean that I can keep calling this function and add multiple interceptors, is that the case?  From looking at the code there appears to only ever be one interceptor.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/main/java/ratpack/http/client/HttpClientSpec.java, line 99 at r1](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KXksqZu_mfR-Rgdi5k1:-KXksqZu_mfR-Rgdi5k2:bicxzb5) ([raw file](https://github.com/ratpack/ratpack/blob/675674c1beb888179002fc5ced0c1f7ce86c9289/ratpack-core/src/main/java/ratpack/http/client/HttpClientSpec.java#L99)):*
> ```Java
>    * @since 1.5
>    */
>   HttpClientSpec responseIntercept(Action<? super ReceivedResponse> interceptor);
> ```

We don't cater for streamed responses by using ReceivedResponse.  Maybe there needs to be some inheritance between ReceivedResponse and StreamedResponse

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/main/java/ratpack/http/client/internal/ContentAggregatingRequestAction.java, line 73 at r1](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KXktjP8Voh02TNt7AKV:-KXktjP8Voh02TNt7AKW:biu0rq7) ([raw file](https://github.com/ratpack/ratpack/blob/675674c1beb888179002fc5ced0c1f7ce86c9289/ratpack-core/src/main/java/ratpack/http/client/internal/ContentAggregatingRequestAction.java#L73)):*
> ```Java
>         });
>         ReceivedResponse receivedResponse = toReceivedResponse(response, content);
>         client.getResponseInterceptor().execute(receivedResponse);
> ```

We need to do the same in ContentStreamingRequestAction

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java, line 162 at r1](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KXku62EzLDcYAhjaM3T:-KXku62Fv9vIEBB3Au75:b-ldobpi) ([raw file](https://github.com/ratpack/ratpack/blob/675674c1beb888179002fc5ced0c1f7ce86c9289/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java#L162)):*
> ```Java
>       s.readTimeout(getReadTimeout());
>       s.requestIntercept(getRequestInterceptor());
>       s.responseIntercept(getResponseInterceptor());
> ```

This seems quite fragile. If we add new properties to the spec it is too easy to forget to set them here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/main/java/ratpack/http/client/HttpClientSpec.java, line 82 at r1](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KXkqqY4kZVcj9sc1ilA:-KXkv7pwTA0LuQgP_GT_:b-31c8x6) ([raw file](https://github.com/ratpack/ratpack/blob/675674c1beb888179002fc5ced0c1f7ce86c9289/ratpack-core/src/main/java/ratpack/http/client/HttpClientSpec.java#L82)):*
<details><summary><i>Previously, rhart (Rus Hart) wrote…</i></summary><blockquote>

I read this to mean that I can keep calling this function and add multiple interceptors, is that the case?  From looking at the code there appears to only ever be one interceptor.
</blockquote></details>

Ignore this, I can see the append to the action now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 24 at r1](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KXkvroQXjVE-WHpQTE2:-KXkvroQXjVE-WHpQTE3:bs5t0jn) ([raw file](https://github.com/ratpack/ratpack/blob/675674c1beb888179002fc5ced0c1f7ce86c9289/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L24)):*
> ```Groovy
> import java.util.concurrent.TimeUnit
> 
> class HttpClientInterceptorSpec extends BaseHttpClientSpec {
> ```

We should add tests for streaming requests too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 109 at r1](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KXkw00q7hneoClEctie:-KXkw00q7hneoClEctif:bcj2hq6) ([raw file](https://github.com/ratpack/ratpack/blob/675674c1beb888179002fc5ced0c1f7ce86c9289/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L109)):*
> ```Groovy
>       all { HttpClient client ->
>         next single(client.copyWith { spec ->
>           spec.requestIntercept { request ->
> ```

We should prove that we can override the response interceptor too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

johnrengelman:
@rhart updated with support for streaming.

johnrengelman:
Note, that this doesn't support executing async code in the response interceptor by default. Should we implement that right away?

rhart:
I think the important thing is to make sure the API doesn't prevent this happening in the future and we iterate on the solution.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-:-KXvuu5xMT4Ye_2LTSG0:b-2tdkj9)*

johnrengelman:
Hmm, this doesn't work for Promises right now, because I'm executing the interceptor off thread. So, I supposed it'll just be easier to fork an execution right now.

ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
```diff
@@ -194,12 +227,14 @@ public HttpClientSpec readTimeout(Duration readTimeout) {
 
   @Override
   public Promise<ReceivedResponse> request(URI uri, final Action<? super RequestSpec> requestConfigurer) {
-    return Promise.async(downstream -> new ContentAggregatingRequestAction(uri, this, 0, Execution.current(), requestConfigurer).connect(downstream));
+    Promise<ReceivedResponse> p =  Promise.async(downstream -> new ContentAggregatingRequestAction(uri, this, 0, Execution.current(), requestConfigurer.append(spec.requestInterceptor)).connect(downstream));
+    return p.next(spec.responseInterceptor);
```
johnrengelman: IDEA wouldn't let me string these 2 lines together. It wouldn't figure out the typing.

johnrengelman:
@ratpack/committers would appreciate some eyes here.

rhart:
*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 66 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgD3D0gKz-L5JrvEtz:-KbgD3D0gKz-L5JrvEu-:b-ux0mem) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L66)):*
> ```Groovy
> 
>   @Timeout(5)
>   def "can override http client"() {
> ```

We're not actually overriding here, we're adding to the existing defined interceptors.  I think I would rename to "can add per request interceptors"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java, line 247 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgIuKsNQok_8f5XcBM:-KbgIuKsNQok_8f5XcBN:bo6f0k) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java#L247)):*
> ```Java
>   private <T extends HttpResponse> Promise<T> intercept(Promise<T> promise, Action<? super HttpResponse> action) {
>     return promise.next(r ->
>       ExecController.require()
> ```

I know we can't mutate the response but I wonder whether we should make the design decision on behalf of users that all response interceptors are run async.  This could be something users decide for themselves and we provide documentation/tests on how to do it.

If we leave it as-is then we should update the javadoc to mention this.  And set the forked execution's event loop to that of the current execution.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 191 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgLRCg3qUasYadVOBl:-KbgLRCg3qUasYadVOBm:b-iqzlto) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L191)):*
> ```Groovy
>           Operation.of {
>             latch.countDown()
>           }.then()
> ```

We could provide an overloaded method on the spec that accepts an operation and wraps it in an action for you that subscribes to it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 86 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgQn7WZx6IPYy7d821:-KbgQn7XBr7sXtM_Ngsl:b-do5wc0) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L86)):*
> ```Groovy
>     handlers {
>       all { HttpClient client ->
>         next single(client.copyWith { spec ->
> ```

I haven't checked how possible this is, just thinking about the api, but what if we were able to add interceptors to the RequestSpec.  So you could do

get { HttpClient client ->
    render client.get(
        otherAppUrl(""), { spec ->
            spec.requestIntercept { request ->
                request.headers { headers ->
                    headers.add 'X-GLOBAL', 'foo'
                }
            }
        })
        .map { it.body.text }
}

Instead of having to use copyWith.  You would still need copyWith to override global interceptors.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

johnrengelman:
Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java, line 247 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgIuKsNQok_8f5XcBM:-KbgTvFl6pElFxMowhGL:b-ytbe2) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java#L247)):*
<details><summary><i>Previously, rhart (Rus Hart) wrote…</i></summary><blockquote>

I know we can't mutate the response but I wonder whether we should make the design decision on behalf of users that all response interceptors are run async.  This could be something users decide for themselves and we provide documentation/tests on how to do it.

If we leave it as-is then we should update the javadoc to mention this.  And set the forked execution's event loop to that of the current execution.
</blockquote></details>

Need @alkemist's input on this one.

---

*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 86 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgQn7WZx6IPYy7d821:-KbgTDkuevzFJ250p1b7:bgyegl2) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L86)):*
<details><summary><i>Previously, rhart (Rus Hart) wrote…</i></summary><blockquote>

I haven't checked how possible this is, just thinking about the api, but what if we were able to add interceptors to the RequestSpec.  So you could do

get { HttpClient client ->
    render client.get(
        otherAppUrl(""), { spec ->
            spec.requestIntercept { request ->
                request.headers { headers ->
                    headers.add 'X-GLOBAL', 'foo'
                }
            }
        })
        .map { it.body.text }
}

Instead of having to use copyWith.  You would still need copyWith to override global interceptors.


</blockquote></details>

Yeah, I'll add that as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 86 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgQn7WZx6IPYy7d821:-KbgUC82BYCIbROs8wDi:bjs28fm) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L86)):*
<details><summary><i>Previously, johnrengelman (John Engelman) wrote…</i></summary><blockquote>

Yeah, I'll add that as well.
</blockquote></details>

In this case we could add a flag I guess to override the globally defined ones?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

johnrengelman:
Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 86 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgQn7WZx6IPYy7d821:-KbgUNMpWpPhTI3uCFIz:b-jvnbln) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L86)):*
<details><summary><i>Previously, rhart (Rus Hart) wrote…</i></summary><blockquote>

In this case we could add a flag I guess to override the globally defined ones?
</blockquote></details>

I'm not sure overriding is helpful. You would simply create a new HttpClient using HttpClient.of

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 86 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgQn7WZx6IPYy7d821:-KbgU_OHt_ly635pw-Dg:b-y4qo48) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L86)):*
<details><summary><i>Previously, johnrengelman (John Engelman) wrote…</i></summary><blockquote>

I'm not sure overriding is helpful. You would simply create a new HttpClient using HttpClient.of
</blockquote></details>

Right, makes sense

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

johnrengelman:
Review status: 0 of 12 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 86 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgQn7WZx6IPYy7d821:-KbgUUxZEdGPirYFm3mC:b-uayrz4) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L86)):*
<details><summary><i>Previously, rhart (Rus Hart) wrote…</i></summary><blockquote>

Right, makes sense
</blockquote></details>

Plus, the method on RequestSpec wouldn't allow you to return a new HttpClient to inject into the chain. 

And actually, now I'm not sure having this on RequestSpec would be useful, since it only applies at the point of defining the client. At that point you can write whatever request/response handling code you want without the need for interceptors.

So I'm thinking this isn't necessary. In the example above, you could simply write:
```
get { HttpClient client ->
  render client.get(
    otherAppUrl(""), { spec ->
      spec.headers { headers ->
        headers.add 'X-GLOBAL', 'foo'
      }
    }
  )
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

rhart:
*[ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy, line 86 at r6](https://reviewable.io:443/reviews/ratpack/ratpack/1124#-KbgQn7WZx6IPYy7d821:-KbgX0Xk4aP4IN89Zk-d:b-ozo7c3) ([raw file](https://github.com/ratpack/ratpack/blob/3ccf0b4e467afc3f220747669f9935efdc1e0f48/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L86)):*
<details><summary><i>Previously, johnrengelman (John Engelman) wrote…</i></summary><blockquote>

Plus, the method on RequestSpec wouldn't allow you to return a new HttpClient to inject into the chain. 

And actually, now I'm not sure having this on RequestSpec would be useful, since it only applies at the point of defining the client. At that point you can write whatever request/response handling code you want without the need for interceptors.

So I'm thinking this isn't necessary. In the example above, you could simply write:
```
get { HttpClient client ->
  render client.get(
    otherAppUrl(""), { spec ->
      spec.headers { headers ->
        headers.add 'X-GLOBAL', 'foo'
      }
    }
  )
}
```
</blockquote></details>

I think where it makes sense is if that logic was encapsulated in an interceptor.  But I see what you mean.  There's not really strong enough an argument to add to RequestSpec.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1124)*

hyleung:
@johnrengelman hey, was this change released? I thought it would be in v1.5.1, but looking at the tag, I' not seeing it (https://github.com/ratpack/ratpack/blob/v1.5.1/ratpack-core/src/main/java/ratpack/http/client/HttpClient.java)

johnrengelman:
It’ll be in 1.6

johnrengelman:
You should be able to get the latest 1.6.0-SNAPSHOT build from here: http://oss.jfrog.org/artifactory/repo/

hyleung:
Ah ic...thanks! That particular url didn't work but this did:

```
maven {
     url "http://oss.jfrog.org/artifactory/oss-snapshot-local"
}
```

PR #1136, opened 2016-12-09
Child Executions

johnrengelman:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1136)


johnrengelman:
@ratpack/committers need some eyes.

tomdcc:
If / when this lands one thing that would be very useful would be the ability to know if a particular execution is a child and who its parent is in an `ExecInitializer` or `ExecInterceptor`. Maybe the provided registry could have a `ParentExecution` reference or something similar, which wouldn't require changing the signature of those interfaces.

johnrengelman:
@tomdcc can you elaborate on how that would be used?

I'm looking at merging this in and then opening additional issues for further use cases.

tomdcc:
@johnrengelman Sure!

I want the ability to add stuff to a child execution's registry based on its parent registry. It wouldn't be exactly the same object, but would be related to it (and would hide the object with the same type from the parent registry).

My specific use case is the [MiniProfiler JVM](https://github.com/tomdcc/miniprofiler-jvm) [Ratpack integration](https://github.com/tomdcc/miniprofiler-jvm/tree/master/miniprofiler-ratpack/src/main/java/io/jdev/miniprofiler/ratpack). The MiniProfiler has the concept of a profiling session called a `Profiler`, and you can also create child profilers for the purposes of displaying forked executions. It keeps track of which `Profiler` is currently active by looking at the instance that is in the current `Execution`'s registry.

The child profiler model was basically designed for forked child executions. The trick is that we don't want the parent's `Profiler` in the child execution's registry, we want a new child of that profiler in there.

I have a [utility method](https://github.com/tomdcc/miniprofiler-jvm/blob/0d9542a4d5b5d6dd12a69276b1da507a0f81910a/miniprofiler-ratpack/src/main/java/io/jdev/miniprofiler/ratpack/MiniProfilerRatpackUtil.java#L82-L121) that does this for normal forked executions (since that's all we've got). It would be nice if instead of having to call it from each child fork point in app code, the logic could just happen inside the normal integration code which [lives in an `ExecInitializer`](https://github.com/tomdcc/miniprofiler-jvm/blob/0d9542a4d5b5d6dd12a69276b1da507a0f81910a/miniprofiler-ratpack/src/main/java/io/jdev/miniprofiler/ratpack/MiniProfilerExecInitializer.java#L82), and app code can just call `forkChild()` without having to know about it. That requires the initializer to be able to tell when it's initializing a child execution and be able to reliably get the direct parent's `Profiler` instance or determine that there isn't one. I don't think just comparing the differences between `Registry.get()` and `Registry.getAll()` will be enough.

johnrengelman:
Hi @tomdcc that's interesting. I think what we are looking for then is some syntax/method that lets us inspect the parent execution as a registry when initializing the child registry.

Something like:
```
ExecStart.forkChild().registryOf(parentRegistry -> Registry.single(parentRegistry.get(Profiler.class).newChild("foo"))
```

What do you think of that? I hesitate to put a reference to the parent execution because that would probably lead to some memory reference issues when trying to garbage collect things. Particularly, if there is a deep nesting of child forking.

tomdcc:
@johnrengelman With a bit of sugar I can see that working nicely when the user does bother doing it, would end up as simple as `ExecStart.forkChild().registryOf(childProfilerOfName("my child execution"))`. Ideally though it would work without any special code at each child fork site (though the user wouldn't get a nice name for the execution).

I get not wanting to have references to the parent execution. An alternative could be either modifying `ExecInitializer` to accept a reference to the parent (execution or registry), or a new interface that is called just when implementing child executions. Another possibility is making the parent registry explicitly available through some method on the execution / registry. Another possibility is providing a way to determine if a particular object is bound to this registry directly or its parent. Any of those would be enough in an initializer to get something working out of the box without manual instrumentation of app code.

johnrengelman:
@ratpack/committers - this is ready to go in with a couple thumbs up.

ldaley:
On it.

ldaley:
@johnrengelman I think we need to take a step back here and define what this is trying to achieve. 

We also have a problem in blindly inheriting the parent's registry in that this can be a memory leak. The parent may have some memory intensive object in its registry and we are now possibly forcing an unnecessary reference to hang around.

I think we should do something like:

1. Introduce `ExecutionRef` (exposes an executions registry as read only)
2. Add `ExecutionRef getParent()` to `Execution`
3. Change the contract of `ExecInitializer` to guarantee that parent executions do not complete until child executions are initialized (i.e. to allow initializers to safely call `getParent()`)

We will probably need to change `SimpleMutableRegistry` to use a concurrent deque instead of a straight list. 

We need to implement `ExecutionRef` in such a way that `Execution` (and its registry) can be GC'd when the execution is finished.

Please let me know if you intend to work on this.

johnrengelman:
@ldaley I'll try and take a swing at that approach this week. I'd love to take a stab and then learn the "right" way to do it ;)

ldaley:
@johnrengelman I've been working on this this morning and am nearly done. I'm going to push as a separate branch/PR and we can do gap analysis from there.

johnrengelman:
ah, sweet!

ldaley:
Superseded by #1311.

PR #1138, opened 2016-12-18
Remove <op>IfOrElse in favor of Function.applyIfOrElse.

johnrengelman:
This is a proof of a new paradigm for the API such that we don't have method explosion for various flow control. Instead we could use a nested function to provide the logic were applicable.
This removes some methods that have been added but not yet released thus there isn't a breaking change. 

Methods that previously existed and provide only a single branch of flow control (e.g. `Promise.mapIf(Predicate, Function)` would remain un-deprecated as it would be difficult to abstract the desired behavior into a single static method on `Function`

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1138)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/exec/Promise.java
+++ b/ratpack-core/src/main/java/ratpack/exec/Promise.java
@@ -539,40 +539,7 @@ public interface Promise<T> {
    * @since 1.4
    */
   default Promise<T> mapIf(Predicate<? super T> predicate, Function<? super T, ? extends T> transformer) {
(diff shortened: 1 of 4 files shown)


alkemist:
@johnrengelman I've taken this idea and reworked it with https://github.com/ratpack/ratpack/commit/725247055bac6808833cfe522419b664fc102e69. Please take a look and let me know what you think.

A next step might be give `Action` the same treatment for use in operation-ish scenarios.

johnrengelman:
Closing this in favor of work in `master`

PR #1142, opened 2016-12-30
Enable SSL SNI support in default http client

### LINKED ISSUE #1077: Handshake Failure when making requests to servers using SNI SSL
Example Workaround: https://gist.github.com/danthegoodman/2b4560e607e760d9895b2c3de10ada01

To properly communicate with a server over SSL that uses SNI, additional information needs to be provided to the SSL engine, specifying the target host a port of the destination server.  When ratpack creates the engine, it does not specify the host and port. 

https://github.com/ratpack/ratpack/blob/5cf0ef6e8417cb59ce922271c8565051dfdc1096/ratpack-core/src/main/java/ratpack/http/client/internal/RequestActionSupport.java#L259

Specifying this information is as easy as calling `createSSLEngine(host,port)` instead of `createSSLEngine()`. The question is 1) can it be applied to every request and 2) how should that information be obtained?

If (1) it can be applied to every request, the ssl handler is already being created uniquely to a uri, so (2) we could pass in the request uri to the `createSslHandler` method.

If (1) it cannot be applied to every request, then (2) we'll need to add some sort of specification to the `RequestSpec` or `HttpClientSpec`.

carhartl:
By creating an SSL engine with peer host and port the http client can support SSL SNI out of the box. Some apis, for instance when provided via AWS API Gateway, may require SNI in order to access them.

Fixes #1077.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1142)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/RequestActionSupport.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/RequestActionSupport.java
@@ -275,9 +275,9 @@ abstract class RequestActionSupport<T> implements Upstream<T> {
   private SslHandler createSslHandler() throws NoSuchAlgorithmException {
     SSLEngine sslEngine;
     if (requestConfig.sslContext != null) {
-      sslEngine = requestConfig.sslContext.createSSLEngine();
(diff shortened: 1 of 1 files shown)


carhartl:
Still thinking about how to test this...

PR #1151, opened 2017-01-25
Fix issue with metrics truncating seconds from request time measurement

jeff-blaisdell:
During a load test on one of Ratpack apps prior to going to production we noticed our Datadog dashboard response times never exceeding 1 second.  Checked out some of our other Ratpack projects and noticed similar behavior.

During debugging it seems Ratpack's metrics module is truncating the seconds and only using the durations nanos for reporting.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1151)


rhart:
Thanks, Jeff!  I've been bitten by this before :(

PR #1158, opened 2017-02-07
Log StartupFailureException.

jterhune:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1158)


beckje01:
Fixed in #1160

PR #1169, opened 2017-02-28
Keep alive header not set

jonmort:
A change between 1.3 and 1.4.0 resulted in the keep-alive header never being set. The change was made here: https://github.com/ratpack/ratpack/commit/184d8804becb4d41641c8817a6e06f05f21ed5c2#diff-d70a1c937265cad0083814db2236676e

Also KeepAliveSupportSpec did not actually assert anything. This change adds some checks for keep alive headers.

I'm a little wary of this change as it changes the behaviour for all responses, but it does appear to be the intended behaviour. AFAIK any HTTP 1.1 requests that don't explicitly close the connection should expect to handle a connection: 'keep-alive' header

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1169)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/server/internal/DefaultResponseTransmitter.java
+++ b/ratpack-core/src/main/java/ratpack/server/internal/DefaultResponseTransmitter.java
@@ -94,6 +94,8 @@ public class DefaultResponseTransmitter implements ResponseTransmitter {
         isKeepAlive = false;
       } else if (!isKeepAlive) {
         responseHeaders.set(HttpHeaderNames.CONNECTION, HttpHeaderValues.CLOSE);
+      } else {
+        responseHeaders.set(HttpHeaderNames.CONNECTION, HttpHeaderValues.KEEP_ALIVE);
       }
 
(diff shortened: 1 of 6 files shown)


lw346:
HTTP/1.0 and HTTP/1.1 have differing default behaviours for Keepalive connections: https://tools.ietf.org/html/rfc7230#section-6.3

For 1.0, connections are considered single-request only by default, and so Connection: keepalive should be sent by the server if it wishes to allow the connection to be persisted.

For 1.1, connections are considered persistent by default, and so Connection: close should be sent by the server if it does *not* wish to allow the connection to be persisted.

The current behaviour is consistent with the spec - although the spec does not preclude the addition of 'Connection: keepalive' by HTTP/1.1 connections, nor the addition on 'Connection: close' by HTTP/1.0 connections.

The trade-off then is on the overhead in adding and transmitting the header vs clients behaving incorrectly?

alkemist:
As @lw346 points out, we are currently following the spec. Closing as wontfix.

PR #1172, opened 2017-03-06
Extends 'by method handler' builder with handler class/instance arguments

timic:
In my use case I have many handler classes. This PR allows me to map them directly to http method like this:

```
handlers {
  prefix "users/:id" {
    byMethod {
      get GetUserHandler
      put PutUserHandler
      delete DeleteUserHandler
    }
  }
}
```

Without these changes I have to write code like this:

```
handlers {
  prefix "users/:id" {
    byMethod {
      get {
        context.insert(context.get(GetUserHandler))
      }
      put {
        // or like this
        context.get(PutUserHandler).handle(context)
      }
      ...
    }
  }
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1172)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/handling/ByMethodSpec.java
+++ b/ratpack-core/src/main/java/ratpack/handling/ByMethodSpec.java
@@ -34,12 +34,44 @@ public interface ByMethodSpec {
   ByMethodSpec get(Block block);
 
   /**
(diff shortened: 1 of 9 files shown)


alkemist:
Thanks @timic , this will be in 1.5.

PR #1184, opened 2017-04-09
Use noneMatch instead of negating anyMatch

kamilszymanski:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1184)


alkemist:
Thanks @kamilszymanski

PR #1204, opened 2017-04-24
fix shell scripts classpath

katoquro:
issue https://github.com/ratpack/ratpack/issues/1203

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1204)

PR #1213, opened 2017-05-10
Use jar archive name as pattern in shell script fix.

katoquro:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1213)


katoquro:
This PR solve the issue more correct than https://github.com/ratpack/ratpack/pull/1204
Please merge it instead

ldaley:
Thanks for the fix!

PR #1224, opened 2017-06-02
ByContentSpec should accept CharSequence for Types

beckje01:
Instead of taking String we want to accept CharSequence for the `type` methods so we can use the Netty Constants

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1224)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/handling/ByContentSpec.java
+++ b/ratpack-core/src/main/java/ratpack/handling/ByContentSpec.java
@@ -78,7 +78,7 @@ public interface ByContentSpec {
    * @param block the code to invoke if the content type matches
    * @return this
    */
-  default ByContentSpec type(String mimeType, Block block) {
+  default ByContentSpec type(CharSequence mimeType, Block block) {
     return type(mimeType, Handlers.of(block));
   }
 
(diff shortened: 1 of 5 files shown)


ldaley:
Unfortunately, this is a binary incompatible change. We'll have to add them as overloads.

beckje01:
I'll get that fixed I thought it was ok since string extended charseq

ldaley:
It's a source compatible change, but not a binary compatible change. Existing code would recompile against it fine without modification, but stuff compiled against the old version would not link to the new one. Dynamic Groovy would still work, but static Groovy and Java won't.

This is the best and most concise reference for what changes are compatible: https://wiki.eclipse.org/Evolving_Java-based_APIs_2

ratpack-core/src/main/java/ratpack/handling/ByContentSpec.java
```diff
@@ -78,7 +78,7 @@
    * @param block the code to invoke if the content type matches
    * @return this
    */
-  default ByContentSpec type(String mimeType, Block block) {
+  default ByContentSpec type(CharSequence mimeType, Block block) {
```
johnrengelman: This works ok because `String` is a `CharSquence`, right? We aren't breaking any compatability?

ldaley: This breaks binary compatibility, but not source compatibility. It needs to be added as an overload.

johnrengelman:
Closing in favor of https://github.com/ratpack/ratpack/pull/1315

PR #1228, opened 2017-06-19
Update DefaultRequestFixture.java

ean5533:
Added `pathBinding` overload that allows caller to set the `PathBinding`'s description.

My team is currently stuck using reflection to set the `description` because there's no other way we can find to set it, and we need it to be a certain value for our unit tests.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1228)


beckje01:
Can you also add this to `RequestFixture.java` With some javadoc?

ean5533:
@beckje01 Whoops, yes, forgot to `git add` that file to the commit. Fixed

ean5533:
Ping -- can I get some eyes on this?

I'm not sure what exactly "50% of diff hit (target 74.7%)" means or what I should do about it.

beckje01:
I'll take a look tonight.

PR #1249, opened 2017-08-07
ISSUE-1244: Expose some metrics on HttpClient

jeff-blaisdell:
Attempt at adding some metric reporting on the HttpClient channel pools.

* Adds new flag to HttpClient to `enable` metric collection.
   * When flag is `false` HttpClient behaves same as today (default behavior)
   * When flag is `true` HttpClient will use instrumented channel pool handlers to track channel pool usage.
* Adds a new `HttpClientMetrics` class to ratpack-dropwizard-metrics to report on channel pool usage.  Default behavior is for it to be disabled.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1249)

### DIFF
--- a/ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/internal/HttpClientMetrics.java
+++ b/ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/internal/HttpClientMetrics.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2013 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ratpack.dropwizard.metrics.internal;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.MetricRegistry;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import ratpack.exec.ExecController;
+import ratpack.http.client.HttpClient;
+import ratpack.http.client.internal.DefaultHttpClient;
+import ratpack.http.client.internal.HttpClientStats;
+import ratpack.service.Service;
+import ratpack.service.StartEvent;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+@Singleton
+public class HttpClientMetrics implements Service, Runnable {
+
+  private static final String METRIC_PREFIX = "httpclient.";
+  private static final String TOTAL_ACTIVE_CONNECTIONS = getMetricName("total.active.connections");
+  private static final String TOTAL_IDLE_CONNECTIONS = getMetricName("total.idle.connections");
+  private static final String TOTAL_CONNECTIONS = getMetricName("total.connections");
+
+  private final HttpClient httpClient;
+  private final MetricRegistry metricRegistry;
+  private final HttpClientConfig config;
+  private final ConcurrentMap<String, HttpMetricGauge> gauges;
+
+  @Inject
+  public HttpClientMetrics(
+    HttpClient httpClient,
+    MetricRegistry metricRegistry
+  ) {
+    boolean enabled = Boolean.getBoolean("ratpack.metrics.httpclient.enabled");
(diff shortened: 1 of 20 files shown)


ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/internal/HttpClientMetrics.java
```diff
@@ -0,0 +1,114 @@
...
+    this.metricRegistry = metricRegistry;
+    this.config = config.getHttpClient();
+    this.gauges = new HashMap<>();
+  }
+
+  @Override
+  public void onStart(StartEvent event) throws Exception {
+    if (config.isEnabled()) {
+      ExecController execController = event.getRegistry().get(ExecController.class);
+      execController.getExecutor().scheduleAtFixedRate(this, 0, config.getPollingFrequencyInSeconds(), TimeUnit.SECONDS);
```
jeff-blaisdell: I don't particular like this polling method of checking the HttpClient stats, but no other ideas forth coming.  It does seem to work, but open to other options.

ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/internal/HttpClientMetrics.java
```diff
@@ -0,0 +1,114 @@
...
+
+  private static String getMetricName(String name) {
+    return METRIC_PREFIX + name;
+  }
+
+  private static String getHostMetricName(String host, String name) {
+    return getMetricName(host + "." + name);
+  }
+
+  private static class HttpMetricGauge implements Gauge<Long> {
```
jeff-blaisdell: Also felt a bit weird making my own gauge.  Couldn't find an implementation that let me just set the value.  Ideas here welcome.

ldaley:
Hi @jeff-blaisdell. Thanks for the great work.

We generally don't add new public API in patch releases. So this change as is, needs to go in 1.6. Alternatively, we could rework this change to make all of the API internal for 1.5.1 (i.e. move new types to the internal package, and move new methods to the implementation type) and then move them into the public space in 1.6.

What would you like to do? Do you need this sooner rather than later, so in 1.5.1?

jeff-blaisdell:
I personally like the idea of soft releasing in 1.5.1.  It will give us a chance to put a few miles on it in our load test and production environments prior to 1.6.

I updated the PR to move all that I could into `internal` packages.  The only place it kind of leaks is in `DropwizardMetricsConfig` as I needed a hook to enable it.  Everything in this PR defaults to disabled, so there shouldn't be any change from current state after merge.

Lastly, I did the refactor to `internal` packages in a separate commit to make it easier to switch back.  Can squash if folks prefer.

jeff-blaisdell:
Just a ping on this one.  Is there anything else that should be addressed prior to merge?

johnrengelman:
Hey @jeff-blaisdell, I think if we're targeting to try and get this in to a 1.5.x patch release, then we'll need to remove the configuration methods in `DropwizardMetricsModule` and instead rely on a static envvar or system property to enable the collection. That would keep everything out of the public API and allow you to do some testing in production by enabling it through those settings and then we can promote it into the public API in 1.6 and expose the configuration methods.

Thoughts @ldaley?

jeff-blaisdell:
I'll see about making the change to a system property today.

jeff-blaisdell:
Resolved merged conflict and removed public facing api from `DropwizardMetricConfig`.  Replaced w/ the following system properties:

```
boolean enabled = Boolean.getBoolean("ratpack.metrics.httpclient.enabled");
int pollingFrequency = Integer.getInteger("ratpack.metrics.httpclient.pollingFrequency", 30);
```

johnrengelman:
Reviewed 9 of 26 files at r1, 15 of 19 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/ratpack/ratpack/1249)*

johnrengelman:
This looks good to go into the 1.5.x branch since there's no public API.

jeff-blaisdell:
Does that mean you want me to re-target this PR against that branch?  Or will Ratpack folks merge to master and cherry pick to branch?

johnrengelman:
we’ll do the merge directly. No action on your part.

johnrengelman:
I've merged this into the `1.5.x` branch.
@ldaley I've added the 1.5.2 milestone to this and will leave it open as a reminder to do a final review before we close out 1.5.2.

johnrengelman:
I also added https://github.com/ratpack/ratpack/issues/1312 to make sure we promote it in 1.6

ldaley:
@jeff-blaisdell I couldn't include this in `1.5.x` sorry.

When @johnrengelman added it to 1.5.x, he merged `master` → `1.5.x` which brought a bunch of unwanted stuff. I only just noticed this. I tried to just cherry pick your changes, but it builds on some stuff in 1.6 that is not in 1.5.

I need to get 1.5.2 out so I am just going ahead with that. What I can offer you is that I will work hard to get 1.6 out soon.

johnrengelman:
Hmm. I wonder how I managed that. I must have turned the brain off when. Doing that. Sorry.

jeff-blaisdell:
Oh well.  It happens.  Thanks for the update.

PR #1251, opened 2017-08-16
Bump hystrix version to 1.5.13

rpalcolea:
There are several release since 1.5.1 -> https://github.com/Netflix/Hystrix/releases

A lot of bugs fixed around semaphores and unsubscribe that could impact Hystrix with RxJava

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1251)


ldaley:
THanks @rpalcolea

PR #1263, opened 2017-08-31
Handling trailing slash in path in DefaultHttpUrlBuilder

jansyk13:
Allows to add paths that contain trailing slash.

This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1263)


jansyk13:
Fixed ourselves, by implementing own url builder

PR #1278, opened 2017-10-07
Adding read/connect timeouts to retrofit, closes #1277

### LINKED ISSUE #1277: Add retrofit read and connect timeouts per service interface

drmaas:
Add read and connect timeouts per service interface.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1278)

### DIFF
--- a/ratpack-retrofit2/src/main/java/ratpack/retrofit/RatpackRetrofit.java
+++ b/ratpack-retrofit2/src/main/java/ratpack/retrofit/RatpackRetrofit.java
@@ -27,6 +27,7 @@ import retrofit2.Retrofit;
 import retrofit2.converter.scalars.ScalarsConverterFactory;
 
 import java.net.URI;
+import java.time.Duration;
 
 /**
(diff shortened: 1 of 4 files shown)


johnrengelman:
@drmaas I'm not sure this is necessary. I think we should instead modify https://github.com/ratpack/ratpack/blob/master/ratpack-retrofit2/src/main/java/ratpack/retrofit/internal/RatpackCallFactory.java#L82 to look for a `HttpClient` in the `Execution` before falling through to the `Context`. Then you can override the value by putting a preconfigured `HttpClient` in the execution.

Additionally, we could make this more natural by adding a method to `RatpackRetrofit.Builder` that takes a `Factory<? extends HttpClient>`, allowing the user to provider their own factory method to generate the `HttpClient` on demand. Then we'd move the look up code from above into the default `Factory` that is defined in the `Builder`.

Thoughts?

johnrengelman:
Closing this in favor of https://github.com/ratpack/ratpack/pull/1306 so we don't have to expose all the `HttpClient` configuration up through the Retrofit API.

PR #1288, opened 2017-11-04
Adding PooledByteBufAllocatorMetricSet for Dropwizard metrics

jansyk13:
JVM memory metric set is not enough for production monitoring(exposes only heap/off heap usage) and there is no way to get more details what Netty does under the hood. This can become handy for example when Ratpack handles large body requests.
Notes:
* adding Dropwizard metric set
* exposing Netty's byte buf allocation metrics
* exposes metrics from PooledByteBufAllocator.DEFAULT because it is used by RatpackDefaultServer
* can expose more detailed metrics(about arenas - chunk of memory allocated using malloc)

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1288)

### DIFF
--- a/ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/DropwizardMetricsConfig.java
+++ b/ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/DropwizardMetricsConfig.java
@@ -33,8 +33,6 @@ public class DropwizardMetricsConfig {
   public static final Duration DEFAULT_INTERVAL = Duration.ofSeconds(30);
 
   private boolean jvmMetrics;
-  private boolean byteBufAllocatorMetrics;
-  private boolean detailedByteBufAllocatorMetrics;
   private boolean requestTimingMetrics = true;
   private boolean blockingTimingMetrics = true;
   private Map<String, String> requestMetricGroups;
@@ -44,6 +42,7 @@ public class DropwizardMetricsConfig {
   private Optional<CsvConfig> csv = Optional.empty();
   private Optional<Slf4jConfig> slf4j = Optional.empty();
   private Optional<GraphiteConfig> graphite = Optional.empty();
+  private Optional<ByteBufAllocatorConfig> byteBufAllocator = Optional.empty();
 
   /**
    * The state of jvm metrics collection.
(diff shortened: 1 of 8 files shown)


ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/DropwizardMetricsModule.java
```diff
@@ -240,6 +241,18 @@ public void onStart(StartEvent event) throws Exception {
...
         metricRegistry.registerAll(new MemoryUsageGaugeSet());
       }
+
+      if (config.isByteBufAllocatorMetrics()) {
+        final MetricRegistry metricRegistry = injector.getInstance(MetricRegistry.class);
+        // Ratpack uses default PooledByteBufAllocator to allocate request bodies and other things.
+        // Allocator is resolved in static block in PooledByteBufAllocator based on OS, presence of Unsafe package, etc.
+        metricRegistry.registerAll(
+          new PooledByteBufAllocatorMetricSet(
+            PooledByteBufAllocator.DEFAULT,
```
johnrengelman: For some parts of Ratpack, the `ByteBufAllocator` is overridable in the Registry - https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/main/java/ratpack/server/internal/ServerRegistry.java#L98

For example: https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/main/java/ratpack/websocket/WebSockets.java#L77

So, we should probably configure metrics collection for both the `DEFAULT` and the `ByteBufAllocator` from the registry, which you can obtain from the `StartEvent`

ratpack-dropwizard-metrics/src/main/java/ratpack/dropwizard/metrics/DropwizardMetricsConfig.java
```diff
@@ -62,6 +64,48 @@ public DropwizardMetricsConfig jvmMetrics(boolean jvmMetrics) {
...
+    return this;
+  }
+
+  /**
+   * The state of detailed byte buf allocator metrics collection. (byte buf allocator metrics needs to be enabled)
+   *
+   * @return True if detailed byte buf allocator metrics collection is enabled. False otherwise
+   * @since 1.6
+   */
+  public boolean isDetailedByteBufAllocatorMetrics() {
```
johnrengelman: If there are multiple configuration options for this feature, then we should create a `ByteBufAllocatorConfig` object and nest the settings together rather than exposing both at the top level of `DropwizardMetricsConfig`.

See how `JmxConfig` is done in this class.

jansyk13:
@johnrengelman Hey I fixed your comments. The bb allocator is now injected from registry, added metric set for unpooled allocator and added config class.

PR #1295, opened 2017-11-17
Upgrade netty to 4.1.17.Final

LukeJoeDavis:
Removed test that tested for content-length on responses that have no content.
This out of spec test is the only one that failed on the upgrade to Netty

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1295)


beckje01:
There are still 5 failing tests for this upgrade https://scans.gradle.com/s/qht2chux7uqzk

ldaley:
That this fails on Linux and not Mac OS points to a regression in Netty with the epoll transport. I'm looking into it.

ldaley:
It turns out that it's a Netty issue: https://github.com/netty/netty/issues/7443

I'm going to workaround this in the tests and put the upgrade in. It's an obscure thing that people are unlikely to hit in a real app.

ldaley:
I've cherry picked this to 1.5.x.

PR #1303, opened 2017-12-19
Return a 404 when client requests a file using a data URL instead of throwing an NPE

craigatk:
We hit a strange corner case on a public-facing site where something (maybe a crawler or bot) was hitting a URL in a data-URL format (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) like `/data:application/json`.

Our app was trying to serve this up as a static file, but `new URI(path).getPath()` in `FileHandler.handle()` was returning `null`, which then got passed to `DefaultFileSystemBinding.file(path)` and that method in turn threw an NPE on `path.startsWith("/")` because `path` was null.

The changes in this PR verify the path and asset are not null before serving the asset and returns a 404 response code if either is null.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1303)


craigatk:
Closing this PR that was targeted at the master branch so I can recreate it pointed at the 1.5.x branch

PR #1304, opened 2017-12-19
Return a 404 when client requests a file using a data URL instead of throwing an NPE

craigatk:
We hit a strange corner case on a public-facing site where something (maybe a crawler or bot) was hitting a URL in a data-URL format (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) like /data:application/json.

Our app was trying to serve this up as a static file, but new URI(path).getPath() in FileHandler.handle() was returning null, which then got passed to DefaultFileSystemBinding.file(path) and that method in turn threw an NPE on path.startsWith("/") because path was null.

The changes in this PR verify the path and asset are not null before serving the asset and returns a 404 response code if either is null.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1304)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/file/internal/DefaultFileSystemBinding.java
+++ b/ratpack-core/src/main/java/ratpack/file/internal/DefaultFileSystemBinding.java
@@ -45,6 +45,10 @@ public class DefaultFileSystemBinding implements FileSystemBinding {
   }
 
   public Path file(String path) {
+    if (path == null) {
+      throw new IllegalArgumentException("Path must not be null");
+    }
+
     if (path.startsWith("/")) {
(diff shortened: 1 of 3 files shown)


ldaley:
The fix is incorrect. The real problem is here: 

https://github.com/ratpack/ratpack/blob/dbec41a973fbbe2d02a09fc8987c341327497961/ratpack-core/src/main/java/ratpack/file/internal/FileHandler.java#L53

The problem is that we parse the URL path as a URL, and this path happens to look like a data URL. The fix is to decode it with something that doesn't try and parse it as a full URL.

Issue #1324, opened 2018-02-17
Add remaining http status codes as const statics to Status class


michaelschlies:
I know this one would simplify some bug detection reports I’m seeing. PR coming tomorrow most likely.

PR #1326, opened 2018-02-17
Added additional HTTP status codes

michaelschlies:
Addresses #1324

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1326)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/Status.java
+++ b/ratpack-core/src/main/java/ratpack/http/Status.java
@@ -25,85 +25,369 @@ import ratpack.http.internal.DefaultStatus;
 public interface Status {
 
   /**
-   * The 1xx status codes.
+   * The 100 status code.
+   * @since 1.6
    */
-
   Status CONTINUE = Status.of(100);
+
+  /**
+   * The 101 status code.
+   * @since 1.6
+   */
   Status SWITCHING_PROTOCOLS = Status.of(101);
+
+  /**
+   * The 102 status code.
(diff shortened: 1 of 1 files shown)


ratpack-core/src/main/java/ratpack/http/Status.java
```diff
@@ -25,16 +25,86 @@
...
+  Status RESET_CONTENT = Status.of(205);
+  Status PARTIAL_CONTENT = Status.of(206);
+  Status MULTI_STATUS = Status.of(207);
+  Status ALREADY_REPORTED = Status.of(208);
+  Status IM_USED = Status.of(226);
 
   /**
-   * The 304 status code.
-   *
-   * @since 1.4
```
johnrengelman: We need to keep the javadoc including the `@since` because this is public API. Each `enum` added here needs to have it's own javadoc comment declaring it is available since 1.6

michaelschlies: Okay, wasn't sure how you guys would want to handle that and figured I'd get feedback on it during the PR. I'll push an update in about 10 minutes.

johnrengelman: No problem. Yeah, any public API needs javadoc declaring the version it became available. Thanks!

michaelschlies:
@johnrengelman Anything else needed for this thing PR?

beckje01:
:+1:

Issue #1316, opened 2018-01-23
Check for null content-type on on Retrofit request


johnrengelman:
When Retrofit uses an empty body (i.e. a POST with no body), then the `request` is non-null, but `contentType()` returns a `null`.

Need to pre-emptively check for the existence of the contentType before trying to pass it on to the Ratpack HttpClient.

https://github.com/ratpack/ratpack/blob/master/ratpack-retrofit2/src/main/java/ratpack/retrofit/internal/RatpackCallFactory.java#L112

beckje01:
Merged in #1338

PR #1334, opened 2018-03-16
Adding pre-emptive null check on Content-Type

collisonchris:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1334)

### DIFF
--- a/ratpack-base/src/test/groovy/ratpack/util/internal/TypeCoercingPropertiesSpec.groovy
+++ b/ratpack-base/src/test/groovy/ratpack/util/internal/TypeCoercingPropertiesSpec.groovy
@@ -121,7 +121,7 @@ class TypeCoercingPropertiesSpec extends Specification {
 
   def "asInetAddress throws an exception if the property value cannot be resolved as a host"() {
     given:
-    delegateProperties.setProperty("key", "null.ratpackframework.org")
+    delegateProperties.setProperty("key", "dev.null.host.com")
 
     when:
     properties.asInetAddress("key")
--- a/ratpack-hystrix/src/main/java/ratpack/hystrix/HystrixMetricsEventStreamHandler.java
+++ b/ratpack-hystrix/src/main/java/ratpack/hystrix/HystrixMetricsEventStreamHandler.java
@@ -40,7 +40,7 @@ import static ratpack.stream.Streams.merge;
  * </pre>
  * <p>
  * This handler can be used in conjunction with Server Sent Event based clients such as the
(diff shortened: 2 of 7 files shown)


collisonchris:
Think this is what you're after in #1316; let me know if any additional updates/tests would be needed

collisonchris:
Looks like I hit a couple of time bombs:

`ratpack.site.LinkCrawlSpec > site has no bad links FAILED
    Condition not satisfied:

    errored.empty
    |       |
    |       false
    https://github.com/Netflix/Hystrix/wiki/Dashboard (referrers: [http://localhost:41459/manual/1.6.0/api/ratpack/hystrix/HystrixMetricsEventStreamHandler.html, http://localhost:41459/manual/1.6.0/hystrix.html, http://localhost:41459/manual/1.6.0/api/ratpack/hystrix/HystrixModule.html], errors: [HTTP status: 404], attempts: 3)
        at ratpack.site.LinkCrawlSpec.site has no bad links(LinkCrawlSpec.groovy:92)`

^this failure is caused by broken links on https://github.com/Netflix; it can be fixed by adding 'https://github.com/Netflix/Hystrix' to the allowBroken list.

`ratpack.util.internal.TypeCoercingPropertiesSpec > asInetAddress throws an exception if the property value cannot be resolved as a host FAILED
    Expected exception of type 'java.net.UnknownHostException', but no exception was thrown
        at org.spockframework.lang.SpecInternals.checkExceptionThrown(SpecInternals.java:79)
        at org.spockframework.lang.SpecInternals.thrownImpl(SpecInternals.java:66)
        at ratpack.util.internal.TypeCoercingPropertiesSpec.asInetAddress throws an exception if the property value cannot be resolved as a host(TypeCoercingPropertiesSpec.groovy:130)`

^this failure is due to 'null.ratpackframework.org' now returning a valid DNS response; dig null.ratpackframework.org returns a valid IP; this can be fixed by changing the bad hostname to something that still returns no entry

Added a commit to address these two

rhart:
I don't think we should add the link to the allowBroken list.  It's not broken, it has permanently been moved.  The link should be changed to `https://github.com/Netflix-Skunkworks/hystrix-dashboard/wiki` and possibly the documentation updated if required.  But that does not have to happen as part of this PR.

collisonchris:
@rhart Makes sense, I'm ok doing the links & doc update and removing the Hystrix from the allowBroken list.

rhart:
Awesome, thanks!

beckje01:
Merged in #1338

PR #1339, opened 2018-04-07
Rx2 Module

beckje01:
Rebased and updated to the latest rx2

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1339)


ratpack-rx2/src/main/java/ratpack/rx2/RxRatpack.java
```diff
@@ -0,0 +1,729 @@
...
+   * <p>
+   * This method must be called during an execution.
+   *
+   * @param observable the observable
+   * @param <T>        the type of the value observed
+   * @return a promise that returns all values from the observable
+   * @throws UnmanagedThreadException if called outside of an execution
+   * @see #promise(Single)
+   */
+  public static <T> Promise<List<T>> promiseAll(Observable<T> observable) throws UnmanagedThreadException {
```
beckje01: Given this does the expensive toList I think its good to have a slightly different name.

PR #1344, opened 2018-04-16
Log cause of failure after decoding HTTP request

jansyk13:
It is quite painful to debug reasons why Netty/Ratpack fails to decode
the HTTP request in production/staging environments. It would be handy
to log request's method and path, but this simple log message should
be enough.

Example would be that Ratpack responds with same status code and response
body for exceeding limit of request line and for exceeding limit for length
of headers.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1344)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/server/internal/NettyHandlerAdapter.java
+++ b/ratpack-core/src/main/java/ratpack/server/internal/NettyHandlerAdapter.java
@@ -114,6 +114,7 @@ public class NettyHandlerAdapter extends ChannelInboundHandlerAdapter {
 
   private void newRequest(ChannelHandlerContext ctx, HttpRequest nettyRequest) throws Exception {
     if (!nettyRequest.decoderResult().isSuccess()) {
+      LOGGER.debug("Failed to decode HTTP request.", nettyRequest.decoderResult().cause());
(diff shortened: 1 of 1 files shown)


ratpack-core/src/main/java/ratpack/server/internal/NettyHandlerAdapter.java
```diff
@@ -114,6 +114,7 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
 
   private void newRequest(ChannelHandlerContext ctx, HttpRequest nettyRequest) throws Exception {
     if (!nettyRequest.decoderResult().isSuccess()) {
+      LOGGER.warn("Failed to decode HTTP request.", nettyRequest.decoderResult().cause());
```
beckje01: Many times a bad request isn't something you want to warn on can we keep this at debug, users will be able to opt in if they want still.

jansyk13: Agree, but I would be fine with warn without stacktrace.

jansyk13:
@beckje01 changed the log level and rebased

To-da:
Any reason why not to merge this one?

beckje01:
Yeah my mistake.

PR #1346, opened 2018-04-17
Don't allow an empty session ID.

johnrengelman:
This prevents the scenario where on logout the session ID is set to "" in the cookie
and a subsequent login transmits "" as the session ID and it is used.

@beckje01 @ldaley  - I've open this against master, but I started it from v1.5.4, so we can release it as a patch version.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1346)


beckje01:
:+1:  Looks correct to me.

Issue #1121, opened 2016-11-28
Request.getPath() should return decoded path


ldaley:
The value returned here should decode any percent escapes.

michaelschlies:
@ldaley based on the conversation over in #1352 does it make sense to revisit this from the perspective of RFC 3986 chapter 2 for this? It does explicitly call out reserved and unreserved characters and provides some guidance surrounding decoding? I'm pretty sure the core of Java's URLDecoder is flawed as it will evaluate any %[\x\x] instances no matter where they appear.

PR #1352, opened 2018-05-03
Request.getPath to return HTTP decoded path

michaelschlies:
Address #1121

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1352)


michaelschlies:
@beckje01 Looks like codecov failed to publish from the Semaphore run, might want to re-trigger a build to publish coverage changes.

beckje01:
Can you add a test covering the encoded path

ldaley:
This is a breaking change, and IIUC has logic flaws. Now, it's impossible to correctly interpret `a%2Fb/c%Fd` because this is returned as `a/b/c/d`, which is semantically different.

I think we should revert this and be clearer about what we are trying to achieve. That's my fault for the very vague issue description.

michaelschlies:
@ldaley I'm on board with that, if you want to shore up the issue (#1121) a bit I'm more than glad to re-implement down another path. I went this route given other places within Ratpack already utilizing the URLDecoder route, so any preferred suggestions would or criteria will drive said implementation.

Thanks!

beckje01:
Reverted at this time we can regroup later

PR #1354, opened 2018-05-09
Fix TypedData.getBytes() sometimes returning Netty's entire buffer

### LINKED ISSUE #1353: TypedData.getBytes() sometimes returns the entire underlying buffer, not accounting for the actual data's offset or length in the buffer
I have a service that relays data from an upstream service to the client and we used code that looked a bit like this (headers and similar details omitted):

```
ctx.get(HttpClient).request(url).then { resp ->
    ctx.response.status(resp.statusCode).send(resp.body.bytes)
}
```

Our upstream service is sending us a `content-length` header and when the response is a big longer, the code worked fine. But for a particular request (or rather, particular arguments), the request is only 1005 bytes long in which case the code did the wrong thing: `resp.body.bytes` returned an array of 16777216 bytes where the first 1005 bytes were indeed our content, but the rest was garbage from previous usages of the buffer. Our code then proceeded to stream the entire buffer contents to the client.

The problem seems to be in `ByteBufBackedTypedData` which checks whether a `ByteBuf` is array backed (which it is, in the failing case) and simply returns `byteBuf.array()`. It should take the buffer length and starting offset into account. I'm using Ratpack 1.5.1.

(sidenote: we're simply passing Netty buffers now and this works perfectly and more efficiently, but the bug still stands)

I can prepare a pull request using [this Stackoverflow answer by a Netty author](https://stackoverflow.com/questions/19296386/netty-java-getting-data-from-bytebuf) for inspiration?

gregopet:
Fixes #1353

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1354)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/internal/ByteBufBackedTypedData.java
+++ b/ratpack-core/src/main/java/ratpack/http/internal/ByteBufBackedTypedData.java
@@ -67,7 +67,10 @@ public class ByteBufBackedTypedData implements TypedData {
   @Override
   public byte[] getBytes() {
     if (byteBuf.hasArray()) {
-      return byteBuf.array();
+      byte[] bytes = new byte[byteBuf.readableBytes()];
+      int readerIndex = byteBuf.readerIndex();
+      byteBuf.getBytes(readerIndex, bytes);
+      return bytes;
     } else {
       ByteArrayOutputStream baos = new ByteArrayOutputStream(byteBuf.writerIndex());
       try {
--- a/ratpack-core/src/test/groovy/ratpack/http/internal/TypedDataSpec.groovy
+++ b/ratpack-core/src/test/groovy/ratpack/http/internal/TypedDataSpec.groovy
@@ -1,0 +1,18 @@
+package ratpack.http.internal
+
+import io.netty.buffer.Unpooled
+import spock.lang.Specification
+
+
+class TypedDataSpec extends Specification {
+
+  def "can return bytes from array-backed Netty buffers"() {
+    given:
+    def buffer = Unpooled.buffer(2000)
+    buffer.writeBytes("foo".bytes)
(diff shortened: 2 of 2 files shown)


ratpack-core/src/main/java/ratpack/http/internal/ByteBufBackedTypedData.java
```diff
@@ -67,7 +67,10 @@ public String getText(Charset charset) {
   @Override
   public byte[] getBytes() {
     if (byteBuf.hasArray()) {
-      return byteBuf.array();
+      byte[] bytes = new byte[byteBuf.readableBytes()];
+      int readerIndex = byteBuf.readerIndex();
+      byteBuf.getBytes(readerIndex, bytes);
```
beckje01: This will be a copy in this case does it make sense to make a new method? We could also make clear that this is the raw array which may be whats wanted in some cases?

gregopet: It does make sense to make it more plain what the method actually does, either in the Javadoc or by changing the method name. And I do see the value of having a copy method because it's easy to mess up the copying if some external code only works with byte arrays. Changing the method name could break some existing code, of course that code had probably worked correctly only by coincidence :)

The current behaviour of returning the raw array itself is probably a bit pointless though because you also need `readerIndex` and `readableBytes` to use it correctly so you might as well work with the `ByteBuf` directly then.

Please advise, I can modify the patch either way.

gregopet: OK - should I mark the old method as `@deprecated` or something? I don't know, I just don't see the reason for it existing as it does now (the other branch of the `if` statement will _still_ make an array copy, so users should either use the `ByteBuf` or explicitly call the new `copyBytes`)?

johnrengelman:
Review for inclusion in 1.6.0

Issue #1027, opened 2016-07-21
Allow executions to have some kind of name/label


alkemist:
Something like Execution.current().name("doing-stuff")

tomdcc:
Would definitely be helpful for the name to be available in an initializer, so that other code that wants to populate its registry in an initializer has access to the name. So should be on `ExecSpec` as well to allow `Execution.fork().name("foo")`

ldaley:
The only thing really stopping us doing this is deciding whether a single opaque string is good enough. Some alternatives might be:

1. Arbitrary object (expected to have a reasonable toString())
2. `Map<String,String>`

I am leaning towards a simple `String` `name` property though.

tomdcc:
Me too. Pretty easy to imagine an arbitrary (and mutable) descriptor object being abused instead of just putting stuff in the registry. Easy enough to call `.name(constructDescriptor('foo', 123, 'bar))` if you need something more structured.

ldaley:
Yeah, let's just settle on a single `CharSequence` property. That leaves the door a little open for someone to do something fancy. I think we should call this property `identifier` or `label`. 

It should go on `ExecSpec` and default to _N_, where that is the _N_th execution forked from that execution. Each `Execution` can maintain a counter for this and `ExecControllerInternal` can provide a counter for top level executions.

Anyone want volunteer to implement?

tomdcc:
I'm happy to but can't promise anything over the next couple of weeks.

PR #1362, opened 2018-05-26
Add execution labels

tomdcc:
Implementation of #1027

### DIFF
--- a/ratpack-exec/src/main/java/ratpack/exec/ExecSpec.java
+++ b/ratpack-exec/src/main/java/ratpack/exec/ExecSpec.java
@@ -126,4 +126,14 @@ public interface ExecSpec {
    */
   ExecSpec eventLoop(EventLoop eventLoop);
 
+  /**
+   * Specifies the initial {@link Execution#getLabel() label} for the execution.
+   * <p>
+   * If neither this method or {@link Execution#label(CharSequence)} are called,
+   * the execution will start with a generated default.
+   *
+   * @param label The label to use for the execution.
+   * @return {@code this}
+   */
+  ExecSpec label(CharSequence label);
 }
--- a/ratpack-exec/src/main/java/ratpack/exec/Execution.java
+++ b/ratpack-exec/src/main/java/ratpack/exec/Execution.java
@@ -197,6 +197,27 @@ public interface Execution extends MutableRegistry {
     return getRef().maybeParent();
   }
 
+
+  /**
+   * The label of this execution.
+   * <p>
+   * This is a human-readable label describing the execution.
+   * It can be set using {@link #label(CharSequence)} or {@link ExecSpec#label(CharSequence)}.
+   * If neither is called, the execution will have a generated default.
+   *
+   * @return the current label of the execution
+   * @since 1.6
+   */
+
+  CharSequence getLabel();
+
+  /**
+   * Set the label of this execution.
+   *
+   * @param label the label
+   */
+  void label(CharSequence label);
+
   /**
    * Registers a closeable that will be closed when the execution completes.
    * <p>
--- a/ratpack-exec/src/main/java/ratpack/exec/internal/DefaultExecController.java
+++ b/ratpack-exec/src/main/java/ratpack/exec/internal/DefaultExecController.java
@@ -29,9 +29,11 @@ import ratpack.registry.RegistrySpec;
 import ratpack.util.internal.ChannelImplDetector;
 import ratpack.util.internal.InternalRatpackError;
 
+import java.util.Optional;
 import java.util.Queue;
 import java.util.concurrent.*;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static ratpack.func.Action.noop;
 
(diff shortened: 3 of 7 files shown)


tomdcc:
In a couple of places I had to change references of the interface type into the concrete type. Another option is introducing an {{ExecutionInternal}} / {{ExecutionRefInternal}} for those cases rather than references to {{DefaultExecution}} / {{DefaultExecution.Ref}}. Happy to do that if it's preferred.

ratpack-exec/src/main/java/ratpack/exec/ExecSpec.java
```diff
@@ -126,4 +126,14 @@
...
   ExecSpec eventLoop(EventLoop eventLoop);
 
+  /**
+   * Specifies the initial {@link Execution#getLabel() label} for the execution.
+   * <p>
+   * If neither this method or {@link Execution#label(CharSequence)} are called,
+   * the execution will start with a generated default.
+   *
+   * @param label The label to use for the execution.
+   * @return {@code this}
```
johnrengelman: Needs an `@since`

ratpack-exec/src/main/java/ratpack/exec/Execution.java
```diff
@@ -197,6 +197,27 @@ default ExecutionRef getParent() {
...
+   * The label of this execution.
+   * <p>
+   * This is a human-readable label describing the execution.
+   * It can be set using {@link #label(CharSequence)} or {@link ExecSpec#label(CharSequence)}.
+   * If neither is called, the execution will have a generated default.
+   *
+   * @return the current label of the execution
+   * @since 1.6
+   */
+
```
johnrengelman: Extra space

ratpack-exec/src/main/java/ratpack/exec/internal/DefaultExecController.java
```diff
@@ -210,6 +221,18 @@ private DefaultExecution createExecution(Action<? super Execution> initialExecut
...
+      private CharSequence getLabel(DefaultExecution.Ref parentRef) {
+        // always call these as we want to increment the counters regardless of whether
+        // an explicit label has been set
+        int forkedExecutionNumber;
+        if (parentRef != null) {
+          forkedExecutionNumber = parentRef.execution.nextChild();
+        } else {
+          forkedExecutionNumber = topLevelExecutionCounter.getAndIncrement();
+        }
+        return maybeLabel.orElse(String.valueOf(forkedExecutionNumber));
```
johnrengelman: so, if I fork 2 executions and each of those forks a child, then the resulting 2 executions will have the same label since we are not combing the parent label with the child's.
Is this the desired behavior? Should we be doing something like `<parent>-<child>` instead?

ratpack-exec/src/main/java/ratpack/exec/internal/DefaultExecution.java
```diff
@@ -631,12 +651,12 @@ void error(Throwable throwable) {
     }
   }
 
-  private static final class Ref implements ExecutionRef {
+  static final class Ref implements ExecutionRef {
```
johnrengelman: This is all a bit to heavy handed just so we can get the `nextChild()` method exposed to the correct place. 
I think it makes more sense to elevate `nextChild()` onto the `ExecutionRef` interface. I'd also probably call it something else (`getAndIncrementChildCount()` is what it is but that's verbose).

johnrengelman:
The parent issue for this PR was assigned to release-1.6.0.
Review this PR for inclusion to push to 2.0.0

michaelschlies:
@johnrengelman I'm working on a pull request to @tomdcc's fork right now addressing your requested changes. Did you have thoughts about the name you'd like to see for `nextChild()` that may be less verbose?

ratpack-exec/src/main/java/ratpack/exec/internal/DefaultExecController.java
michaelschlies: If you had nested forking couldn't you still do this with a <parent>-<child> model? It is entirely possible with (clumsy design most likely) to yield something vaguely like this:

image

ratpack-exec/src/main/java/ratpack/exec/Execution.java
michaelschlies: Removed in my PR against @tomdcc's fork

ratpack-exec/src/main/java/ratpack/exec/ExecSpec.java
michaelschlies: Added in my PR against @tomdcc's fork.

ratpack-exec/src/main/java/ratpack/exec/internal/DefaultExecution.java
michaelschlies: Removed in my PR against @tomdcc's fork

michaelschlies: I did leave this as `nextChild()` for the moment, but would certainly be open to whether you just want me to refactor it to `getAndIncrementChildCount()`

michaelschlies:
@johnrengelman With how long this has been sitting as-is on @tomdcc's fork, would you like me to rebase my PR to his fork and just PR directly to close out #1027 and remove this one from 2.0 milestone?

PR #1365, opened 2018-06-10
Upgraded gradle-wrapper from 2.9 to 4.5 in lazybones template

### LINKED ISSUE #1364: Update gradle-wrapper in lazybones template from 2.9 to 4.5
Ratpack project uses `gradle-wrapper` version 4.5, lazybones creates a project with version 2.9 of gradle-wrapper. I guess there is no reason to use gradle-wrapper 2.9 instead of 4.x so most of the people updates it anyway after creating a new project from the template. Updating it at the lazybones template level could save the time.

wololock:
Fixes #1364 

`gradle-wrapper.jar` file used in this pull request is the same file as the one used for the root project.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1365)


beckje01:
Thanks

PR #1373, opened 2018-07-14
enhance request fixture with multipart form and file upload capabilities

guspower:
Enhancements to RequestFixture to support multipart requests including file uploads.

The Content-Transfer-Encoding will probably need some further elaboration if there proves to be wider need of this testing feature.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1373)

### DIFF
--- a/ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
+++ b/ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
@@ -1,0 +1,396 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ratpack.test.http.internal;
+
+import ratpack.test.http.MultipartFileSpec;
+import ratpack.test.http.MultipartFormSpec;
+
+import java.io.UnsupportedEncodingException;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/*
+    As defined in RFC 1867 and 2388 (see https://tools.ietf.org/html/rfc1867)
+ */
+public class DefaultMultipartForm {
+
+  private final Data data;
+  private final Writer writer;
+  private final Map<String, String> _headers;
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  private DefaultMultipartForm(Data data) {
+    this.data = data;
+    this.writer = new Writer(data);
+    this._headers = new HashMap<>();
+    _headers.put("Content-Type", getContentType());
+  }
+
+  public String getBody() {
+    return writer.write();
+  }
+
+  public String getContentType() {
+    return String.format("multipart/form-data; boundary=%s", data.boundary);
+  }
+
+  public Map<String, String> getHeaders() {
+    return Collections.unmodifiableMap(_headers);
+  }
+
+  final static class Writer {
+
+    private final Data data;
+
+    Writer(Data data) {
+      this.data = data;
+    }
+
+    String write() {
+      StringBuilder result = new StringBuilder();
+
+      addFields(data.fields, data.boundary, result);
+      addFiles(data.files, data.boundary, result);
+      closeBoundary(data.boundary, result);
+
+      return result.toString();
+    }
+
+    private void addFields(Map<String, Set<String>> fields, String boundary, StringBuilder out) {
+      fields.forEach((name, values) -> {
+        values.forEach((value) -> {
+          openBoundary(boundary, out);
+          declareField(name, out);
+          data(value.getBytes(), out);
+        });
+      });
+    }
+
+    private void addFiles(Map<String, Set<File>> files, String boundary, StringBuilder out) {
+      files.forEach((name, filesByField) -> {
+        openBoundary(boundary, out);
+        if (filesByField.size() > 1) {
+          addMultipleFilesToField(filesByField, name, out);
+        } else {
+          addSingleFileToField(filesByField, out);
+        }
+      });
+    }
+
+    private void addMultipleFilesToField(Set<File> files, String name, StringBuilder out) {
+      String subBoundary = generateSubBoundary(data.boundary, name);
+
+      declareField(name, out);
+      type(String.format("multipart/mixed, boundary=%s\n", subBoundary), out);
+      files.forEach((file) -> {
+        openBoundary(subBoundary, out);
+        disposition(String.format("attachment; filename=\"%s\"", file.getName()), out);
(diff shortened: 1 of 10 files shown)


beckje01:
Merged by hand in https://github.com/ratpack/ratpack/commit/572dee54ad8c1c2920edbaa7406a72528687c54c Thanks for the contribution

guspower:
excellent. was there a reason why you had to merge it in by hand?

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
```diff
@@ -0,0 +1,396 @@
...
+
+  public String getContentType() {
+    return String.format("multipart/form-data; boundary=%s", data.boundary);
+  }
+
+  public Map<String, String> getHeaders() {
+    return Collections.unmodifiableMap(_headers);
+  }
+
+  final static class Writer {
```
erdi: Boundry is being passed around quite a lot, but it is immutable from `Data` and could probably be extracted in the constructor and made field in this class which would simplify a number of signatures.

I'm also not a fan of passing the `StringBuilder` instance around and would probably pass it in the constructor and make `write()` a single use method, but that's probably a matter of personal preference.

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
```diff
@@ -0,0 +1,396 @@
...
+      this.fields = Collections.unmodifiableMap(fields);
+      this.files = Collections.unmodifiableMap(files);
+    }
+
+  }
+
+  public final static class Builder implements MultipartFormSpec {
+
+    private final String _boundary;
+    private final Map<String, Set<String>> _fields;
```
erdi: Why is this a map to a set? Is there anything in the spec for `multipart/from-data` that prevents you from having the same value multiple times for a single field name? Because this implementation does.

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
```diff
@@ -0,0 +1,396 @@
...
+    private final Map<String, Set<File>> _files;
```
erdi: Again, not sure what the rationale is behind making this a map to a set and not to a list is.

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
```diff
@@ -0,0 +1,396 @@
...
+
+    Data(String boundary, Map<String, Set<String>> fields, Map<String, Set<File>> files) {
+      this.boundary = boundary;
+      this.fields = Collections.unmodifiableMap(fields);
+      this.files = Collections.unmodifiableMap(files);
+    }
+
+  }
+
+  public final static class Builder implements MultipartFormSpec {
```
erdi: Why are there underscores in the names of fields in this class? I don't think such naming convention is used in ratpack's codebase.

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
```diff
@@ -0,0 +1,396 @@
...
+    public Map<String, Set<File>> getFiles() {
+      return Collections.unmodifiableMap(_files);
+    }
+
+    public DefaultMultipartForm build() {
+      return new DefaultMultipartForm(new Data(_boundary, _fields, _files));
+    }
+
+    public Builder field(String name, String value) {
+      if (!_fields.containsKey(name)) {
```
erdi: Could probably be simplified using `Map.computeIfAbsent()`

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
```diff
@@ -0,0 +1,396 @@
...
+      return this;
+    }
+
+    public FileBuilder file() {
+      return new FileBuilder(this);
+    }
+
+    private Builder add(File file) {
+      String name = file.getField();
+      if (!_files.containsKey(name)) {
```
erdi: Could probably be simplified using `Map.computeIfAbsent()`

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
```diff
@@ -0,0 +1,396 @@
...
+      return this;
+    }
+
+    private static String generateBoundary() {
+      return String.format("TEST%d", System.currentTimeMillis());
+    }
+
+  }
+
+  public final static class FileBuilder implements MultipartFileSpec {
```
erdi: Why are there underscores in the names of fields in this class? I don't think such naming convention is used in ratpack's codebase.

erdi: I don't believe it's used outside of this class so should probably be private.

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
```diff
@@ -0,0 +1,396 @@
...
+
+    public FileBuilder name(String name) {
+      this._name = name;
+
+      return this;
+    }
+
+  }
+
+  public final static class File implements Comparable<File> {
```
erdi: I don't believe it's used outside of this class so should probably be private.

erdi: Why are there underscores in the names of fields in this class? I don't think such naming convention is used in ratpack's codebase.

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
```diff
@@ -0,0 +1,396 @@
...
+      return value != null && !value.isEmpty();
+    }
+
+    private boolean isValueSet(byte[] value) {
+      return value != null && !(value.length == 0);
+    }
+
+  }
+
+  public final static class Encoding {
```
erdi: I don't believe it's used outside of this class so should probably be private.

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultMultipartForm.java
```diff
@@ -0,0 +1,396 @@
...
+
+    final static List<String> VALUES = Stream.of("BASE64", "QUOTED-PRINTABLE", "8BIT", "7BIT", "BINARY")
```
erdi: I believe the idiomatic way of doing this in Java is `Arrays.asList()`

erdi: Should probably be private

ratpack-test/src/test/groovy/ratpack/test/http/DefaultMultipartFormSpec.groovy
```diff
@@ -0,0 +1,237 @@
...
+Content-Disposition: form-data; name="${field.key}"
+
+${field.value}
+--${boundary}--
+"""
+
+        where:
+            builder = DefaultMultipartForm.builder()
+            boundary = builder.boundary
+            field = ['name': 'Werner Heisenberg'].find { true }
```
erdi: Would be more comprehensible as `['name': 'Werner Heisenberg'].entrySet().first()`, IMHO. Took me a while to understand the intention here.

guspower:
thanks for the review @erdi , I will address these comments

PR #1380, opened 2018-08-08
Content Negotiation: Add UTF-8 as default charset for text mime types

guylabs:


johnrengelman:
Review for inclusion in 1.6.0

beckje01:
@guylabs do you mind detailing more how this came up its all looking correct to me, but I would like to be able to give more details in release notes

Issue #975, opened 2016-05-03
Add support for HttpClient request/response interceptors


hyleung:
Ref: https://github.com/ratpack/ratpack/issues/913#issuecomment-214787835 (Zipkin integration)

As far as I can tell, there currently aren't any "hooks" that allow interception of `HttpClient` requests and responses. On the response end of things, I believe it's possible to use [`Promise.wiretap`](https://ratpack.io/manual/current/api/ratpack/exec/Promise.html#wiretap-ratpack.func.Action-) to intercept the response. Don't believe there's a straightforward way to do this for the request.

In my specific scenario, this would make logging of Zipkin client spans easier, but could be more generally useful.

hyleung:
Would it be possible to look for interceptors via the registry? Something like this: https://github.com/hyleung/ratpack/commit/80b6ba1050332b23215927eac42b5837b14fbcc3 (wanted to write the corresponding unit tests to verify it, but couldn't quite figure out how to add entries into the registry in the test setup).

rhart:
That's what I was thinking too.  You can see test examples in the health check code https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/test/groovy/ratpack/health/HealthCheckHandlerSpec.groovy#L102

johnrengelman:
While implementing retrofit integration (https://github.com/ratpack/ratpack/pull/976), I've also realized that interceptors would extremely helpful.
The issue here is if we abstract away the `HttpClient` then we need a way to configure per request type behavior.
My use case was that I wanted to add a specific header to every outgoing request for auth.
Using a Retrofit interface, I could do this, but then would have to declare each and every item like this as part of the API interface.

It would greatly simplifying things if I could globally configure this type of behavior.

Technically speaking, I don't think this would be difficult to implement, but there are some areas around usability that we want to be mindful of.

For example, there is only 1 `HttpClient` registered in the app, so any interceptors would be global. This means, it would be the developer's responsibility to handle any conditional logic (i.e. add a header to requests going to foo.com but not bar.com).
An alternative might be to add an API that does a type of "shallow" clone of `HttpClient` that interceptors could be added to. Then the interceptors are applied only when calling through the "clone".

hyleung:
[...]

PR #1381, opened 2018-08-26
Add errorIntercept method to HttpClientSpec.

lw346:
This method is analagous to the existing responseIntercept method, but for requests that have not resulted in a response, eg. connection refused, connection and read timeouts, invalid responses, etc.

Extends the functionality added in https://github.com/ratpack/ratpack/commit/b9bd8cf1807d5b12d5b389438813f485615f0e13 as part of https://github.com/ratpack/ratpack/issues/975.

For tracing needs, we typically need to close the client-side trace (and populate it with any error details) in the event of unexpected failure as well as the expected response.

See https://github.com/lw346/ratpack-opentracing/blob/49bfb54baec68731bb0fd6c66ce8b467ca31de41/src/main/java/uk/me/lwood/ratpack/opentracing/client/SpanPropagationHandler.java#L61 for an example of use in an OpenTracing canonical style.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1381)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
@@ -301,9 +301,15 @@ public class DefaultHttpClient implements HttpClientInternal {
   }
 
   private <T extends HttpResponse> Promise<T> intercept(Promise<T> promise, Action<? super HttpResponse> action, Action<? super Throwable> errorAction) {
-    return promise.onError(Exception.class, t -> {
-        errorAction.execute(t);
-        throw t;
+    return promise.wiretap(r -> {
+        if (r.isError()) {
+          ExecController.require()
+            .fork()
+            .eventLoop(Execution.current().getEventLoop())
+            .start(e ->
+              errorAction.execute(r.getThrowable())
+            );
+        }
       })
       .next(r ->
         ExecController.require()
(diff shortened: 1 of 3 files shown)


ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
```diff
@@ -279,27 +286,33 @@ public HttpClientSpec enableMetricsCollection(boolean enableMetricsCollection) {
...
-    return promise.next(r ->
-      ExecController.require()
-        .fork()
-        .eventLoop(Execution.current().getEventLoop())
-        .start(e ->
-          action.execute(r)
-        )
-    );
+  private <T extends HttpResponse> Promise<T> intercept(Promise<T> promise, Action<? super HttpResponse> action, Action<? super Throwable> errorAction) {
+    return promise.onError(Exception.class, t -> {
```
lw346: I don't like the use of `Exception` here, but `Action` only allows `Exception` rather than `Throwable` to be thrown, even though `#onError` catches `Throwable`.

johnrengelman: You can use `Promise.wireTap(Action<Result>)` in order to register an error action that is not inline with the promise chain execution. This will allow you to not have to handle and re-throw the Throwable.

We probably should also execute the `errorAction` in a forked execution just like with do with the response interceptor. That way you can add blocking code without impacting the main program flow.

ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy
```diff
@@ -62,6 +62,33 @@ class HttpClientInterceptorSpec extends BaseHttpClientSpec {
     result == 'foo'
   }
 
+  @Timeout(5)
+  def "can intercept errors"() {
```
johnrengelman: Make sure we add tests for verify that we can execute `Promise`s inside the `errorAction` just like we do with the response interceptor below (line 172)

PR #1382, opened 2018-08-27
Add ratpack-gson module

johnrengelman:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1382)


beckje01:
:+1:

johnrengelman:
I'm going to put this into milestone-1.6 for now so we can decide it it ships or not.
I'm ok with it not going into 1.6 if we want to hold off.

PR #1386, opened 2018-09-17
Minor changes to support Java 9

johnrengelman:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1386)


johnrengelman:
Adding to the release-1.6.0 milestone so we can review this. These changes are backwards compatible.

ratpack-spring-boot/src/main/java/ratpack/spring/config/RatpackConfiguration.java
```diff
@@ -51,8 +51,8 @@ public void run(String... args) throws Exception {
     server.start();
   }
 
-  @PreDestroy
-  public void stop() throws Exception {
```
beckje01: Do we need to leave the stop() method since it was public and have it just call destroy if called but we can remove the predestroy?

beckje01:
Looks good I think there is one thing to adjust for binary compatibility

ratpack-spring-boot/src/main/java/ratpack/spring/config/RatpackConfiguration.java
johnrengelman: Ah, that's a good point. Yeah we should probably do that.

Issue #1389, opened 2018-10-02
Support micrometer metrics


drmaas:
This would allow the use of different metrics backends, like dropwizard. https://micrometer.io/

drmaas:
This involves adding a new module, ratpack-micrometer, and adding some new common configs:

* global tags
* NamingConvention
 * MeterFilter(s)

And then also adding new registry configs for those supported by micrometer:

* Atlas
* Datadog
* Ganglia
* Updated graphite
* Influx
* Updated JMX
* New Relic
* Prometheus
* SignalFX
* StatsD
* Wavefront

And then updating existing reporters for:

* console
* slf4j
* websockets
* csv

johnrengelman:
Yeah, we should totally have a module for this.

michaelschlies:
@drmaas @johnrengelman @mustaine @beckje01 Are we thinking we'd like to expand on @mustaine's work to support more backends for this module at this point?

michaelschlies:
^ Oops, missed the revert on the original PR

PR #1404, opened 2018-11-02
Ratpack micrometer #1389

mustaine:
~WIP

I have started ratpack-micrometer using prometheus. #1389
I based some abstraction on ratpack-dropwizard.

Let's discuss implementation details from here?

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1404)


drmaas:
@johnrengelman fyi

beckje01:
This is really cool, I'm wondering if it makes sense to release this in its own repo first get a few integrations working then bring it into the main repo? Doesn't seem to depend on changes in core etc

ratpack-core/src/main/java/ratpack/handling/HandlerDecorator.java
```diff
@@ -163,7 +163,7 @@ static HandlerDecorator noop() {
    * As the {@code rest} argument encapsulates the application handlers, the returned handler should generally delegate to it (via {@link Context#insert(Handler...)}).
    *
    * @param serverRegistry the server registry
-   * @param rest the rest of the handlers of the application
+   * @param rest           the rest of the handlers of the application
```
drmaas: formatting?

johnrengelman:
@mustaine can you rebase this PR to the latest master?

jkwatson:
Seems like it would be better to implement via opencensus, rather than a specific metric consumer. https://opencensus.io/

drmaas:
@jkwatson that was my main point - micrometer itself is capable of reporting dimensional metrics to many different backends (telegraf and influxdb for example), not just prometheus.

jkwatson:
@drmaas yep. Just wanted to point out that opencensus seems to be heading toward being a pretty good standard library to use.  👍

beckje01:
This was merged in #1439

PR #1405, opened 2018-11-03
Provide a more powerful retry mechanism for the Promise API.

felipefzdz:
This PR extends the existing retry API on `Promise` type to support more strategies than the attempts based current one.

This is achieved by providing a couple of new strategies: `RetryPolicy` and `Delay`. The former defines the exhaustion strategy, or when the retries should stop. The latter defines the delay between those retries, enabling the users the ability to provide their own backoff mechanisms.

Existing APIs have been properly deprecated.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1405)

PR #1431, opened 2019-01-02
Provide Powerful Retries for Promises

beckje01:
Rebase and Version changes for https://github.com/ratpack/ratpack/pull/1405

This PR extends the existing retry API on Promise type to support more strategies than the attempts based current one.

This is achieved by providing a couple of new strategies: RetryPolicy and Delay. The former defines the exhaustion strategy, or when the retries should stop. The latter defines the delay between those retries, enabling the users the ability to provide their own backoff mechanisms.

Existing APIs have been properly deprecated.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1431)


johnrengelman:
@beckje01 can you rebase this so we get new CI?

PR #1433, opened 2019-01-07
ParallelBatch support executing code at start of each forked execution

### LINKED ISSUE #1432: Add method to ParallelBatch for executing code in each forked execution before work has been done.
### Problem
When ParallelBatch forks executions, the new executions lose information that the parent execution had.  Most of these problems can be solved using the execInit method (e.g. adding the Context onto the Execution).  However, for setup that needs to be run inside the forked Execution (e.g. [copying over MDC logging values](https://github.com/ratpack/ratpack/blob/05c20b1976a6e8cc4557dc2b101aa28d41690dcf/ratpack-core/src/main/java/ratpack/logging/MDCInterceptor.java#L97-L99)), no clean solution exists (`execInit` runs in parent Execution).

### Proposal
Add a new method to ParallelBatch called `onExecStart` that behaves similarly to `execInit`, but instead runs a code block at the start of each Promise running in it's own forked execution.

```java
  /**
   * Specifies a block of code that will be executed in each forked execution before any work has been done.
   * <p>
   * This can be used to copy over values to the forked execution (e.g. the MDC logging context).
   *
   * @param onExecStart the function to execute
   * @return a new batch, configured to use the given onExecStart block
   */
  ParallelBatch<T> onExecStart(Block onExecStart);
```

tsaude:
This PR adds the `onExecStart` method to `ParallelBatch`, allowing a `Block` to be executed inside each new forked Execution before any work has been done on that Execution.

Closes #1432 

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1433)

### DIFF
--- a/ratpack-exec/src/main/java/ratpack/exec/util/ParallelBatch.java
+++ b/ratpack-exec/src/main/java/ratpack/exec/util/ParallelBatch.java
@@ -23,6 +23,7 @@ import ratpack.exec.Promise;
 import ratpack.exec.util.internal.DefaultParallelBatch;
 import ratpack.func.Action;
 import ratpack.func.BiAction;
+import ratpack.func.Block;
 import ratpack.stream.TransformablePublisher;
 
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


johnrengelman:
Related to https://github.com/ratpack/ratpack/issues/1432.

PR #1437, opened 2019-02-12
Fix relative path redirects urlencoding query parameters, fixes #1436

### LINKED ISSUE #1436: Following URL redirect url-encodes the query parameters resulting in a bad request
In the ratpack HttpClient implementation when we get a 301 response, we read the `Location` header as a full raw string. Then when creating the new URI we are url encoding the full string, which results in the `?` being percent encoded. 

The bug specifically happens here https://github.com/ratpack/ratpack/blob/141fa01f4e42781cf2120ddb53a16ba16fced69a/ratpack-core/src/main/java/ratpack/http/client/internal/RequestActionSupport.java#L258

To fix this we would need to parse the `Location` header to at least separate out the query parameters and pass them to the URI constructor.

fromanator:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1437)

### DIFF
--- a/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientRedirectionSpec.groovy
+++ b/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientRedirectionSpec.groovy
@@ -53,6 +53,73 @@ class HttpClientRedirectionSpec extends BaseHttpClientSpec {
     pooled << [true, false]
   }
 
+  def "can follow redirect get request with query parameters"() {
+    given:
+    bindings {
+      bindInstance(HttpClient, HttpClient.of { it.poolSize(pooled ? 8 : 0) })
+    }
+    otherApp {
+      get("foo2") {
+        redirect(301, otherAppUrl("foo?key1=value1&key2=value2").toString())
+      }
+
+      get("foo") {
+        render context.request.queryParams["key1"] + context.request.queryParams["key2"]
+      }
+    }
+
+    when:
+    handlers {
+      get { HttpClient httpClient ->
+        httpClient.get(otherAppUrl("foo2")) {
+        } then {
+          render it.body.text
+        }
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


ratpack-core/src/test/groovy/ratpack/http/client/HttpClientRedirectionSpec.groovy
```diff
@@ -53,6 +53,73 @@ class HttpClientRedirectionSpec extends BaseHttpClientSpec {
...
+    }
+
+    then:
+    text == "value1value2"
+
+    where:
+    pooled << [true, false]
+  }
+
+  def "can follow a relative redirect get request with query parameters"() {
```
fromanator: Confirmed that with the code without my fix would break here, the application would return a 404 because the `?` was being percent encoded so path matching would break. After my fix this does work.

ratpack-core/src/test/groovy/ratpack/http/client/HttpClientRedirectionSpec.groovy
```diff
@@ -53,6 +53,73 @@ class HttpClientRedirectionSpec extends BaseHttpClientSpec {
     pooled << [true, false]
   }
 
+  def "can follow redirect get request with query parameters"() {
```
fromanator: Added just to be sure that the changes I made for the absolute URI didn't break anything, old code also passed this test.

beckje01: :+1:

ratpack-core/src/test/groovy/ratpack/http/client/HttpClientRedirectionSpec.groovy
beckje01: :+1:

PR #1438, opened 2019-03-29
Upgrade SpringBoot to 2.1.3

beckje01:
Upgrade SpringBoot to 2.1.3 this mostly forces moving to newer testing setup.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1438)

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -35,7 +35,7 @@ ext {
     newrelic         : "3.15.0",
     reactiveStreams  : "1.0.2",
     guice            : "4.1.0",
-    springBoot       : "1.3.0.RELEASE",
+    springBoot       : "2.1.3.RELEASE",
     caffeine         : "2.6.2",
     retrofit         : "2.4.0",
     commonsLogging   : "1.2",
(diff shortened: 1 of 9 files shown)


gradle/dependencies.gradle
```diff
@@ -35,7 +35,7 @@ ext {
     newrelic         : "3.15.0",
     reactiveStreams  : "1.0.2",
     guice            : "4.1.0",
-    springBoot       : "1.3.0.RELEASE",
+    springBoot       : "2.1.3.RELEASE",
```
johnrengelman: Wouldn't this be considered a breaking change? it will force Spring Boot 2.x on users of the module in a minor version of Ratpack.

beckje01:
I think I'm gonna split this into a new module so we don't force an upgrade to boot2 unless people opt into it.

gradle/dependencies.gradle
michaelschlies: Yes, it is. I tried working around that with the java 9 stuff seeing if I could do it in 1.x before and Spring was one of those things I simply couldn't move forward.

johnrengelman:
👍 I think that's the right decision @beckje01

johnrengelman:
Updated to 2.6.3 in ec736cc35aaee69ff5a7d821d10904787f3454df.
Did not break this out as it will be part of 2.x and we'll drop support for 1.x in the major version change.

PR #1446, opened 2019-04-22
Fix ClientSide decryption bug causing intermittent test failures in ClientSideSessionSpec

kristinclemens:
ClientSide's default crypto decryption trims 0-byte padding without regard for the number of remaining bytes, resulting in an IndexOutOfBoundsException when trying to decrypt a lastAccessToken ending with one or more non-padding 0-bytes.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1446)

### DIFF
--- a/ratpack-session/src/test/groovy/ratpack/session/clientside/ClientSideSessionSpec.groovy
+++ b/ratpack-session/src/test/groovy/ratpack/session/clientside/ClientSideSessionSpec.groovy
@@ -47,8 +47,7 @@ class ClientSideSessionSpec extends SessionSpec {
     "DESede/CBC/NoPadding",
     "DESede/CBC/PKCS5Padding",
     "DESede/ECB/NoPadding",
-    "DESede/ECB/PKCS5Padding",
-    "DESede/CBC/NoPadding"
+    "DESede/ECB/PKCS5Padding"
   ]
 
   static int keyLength(String algorithm) {
--- a/ratpack-session/src/test/groovy/ratpack/session/clientside/ClientSideSessionSpec.groovy
+++ b/ratpack-session/src/test/groovy/ratpack/session/clientside/ClientSideSessionSpec.groovy
@@ -25,7 +25,6 @@ import ratpack.session.Session
 import ratpack.session.SessionId
 import ratpack.session.SessionModule
 import ratpack.session.SessionSpec
-import spock.lang.Issue
 
 import java.time.Duration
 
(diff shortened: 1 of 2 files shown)


ratpack-session/src/test/groovy/ratpack/session/clientside/ClientSideSessionSpec.groovy
```diff
@@ -337,62 +343,14 @@ class ClientSideSessionSpec extends SessionSpec {
 
     expect:
     text == "null"
-    postText("set/foo") == "ok"
+    result.set(postText("set/foo"))
```
ldaley: This shouldn't be necessary. postText() is already synchronous.

kristinclemens:
Closing to rework this due to an oversight found during additional testing.

kristinclemens:
Determined the actual issue, which is DefaultCrypto being unaware of whether ending 0-bytes are padding or not, and thus trimming them. This caused an issue when a lastAccessToken ended in one or more non-padding zero-bytes, because `ByteBuf.readLong` in `isValid` would fail due to the writerIndex (as set during decryption) being less than 8. In that case, using `getLong` from the start of the long is sufficient and results in the correct lastAccessToken being read based on extensive testing.

It might make sense to test for this specific case (though it is fairly uncommon in the grand scheme of things). From what I can tell, we'd need to mock `ByteBuf.writeLong` to set a lastAccessToken with a long ending in a 0-byte. I'm not super familiar with Spock mocks, but I could manage if we think it's necessary for this PR.

ratpack-session/src/test/groovy/ratpack/session/clientside/ClientSideSessionSpec.groovy
kristinclemens: Removed!

beckje01:
Can you squash and rebase this?

ratpack-session/src/test/groovy/ratpack/session/clientside/ClientSideSessionSpec.groovy
```diff
@@ -48,7 +47,8 @@ class ClientSideSessionSpec extends SessionSpec {
     "DESede/CBC/NoPadding",
     "DESede/CBC/PKCS5Padding",
     "DESede/ECB/NoPadding",
-    "DESede/ECB/PKCS5Padding"
+    "DESede/ECB/PKCS5Padding",
+    "DESede/CBC/NoPadding"
```
johnrengelman: So, this algorithm is already in the list above. Line 47.

kristinclemens: Ahh, good find. I thought since we were testing it separately in the annotated test that we weren't also testing it above.

Issue #1369, opened 2018-06-16
Race condition in CachingUpstream<T>


benweint:
I've encountered an issue when using `Promise.cache()` with multiple subscribers that are then put into a `ParallelBatch`, where awaiting the results of the `ParallelBatch` sometimes triggers a `NullPointerException` within `CachingUpstream.yield`.

I'm running Ratpack 1.5.4.

Here's a stripped-down test case that reliably reproduces the problem for me (when run in a loop 10k times, I get at least 1 failure consistently):

```
@Test
public void testPromiseCacheWithMultipleSubscribers() throws Exception {
    List<Promise<Integer>> promises = new ArrayList<>();

    for (int i = 0; i < 25; i++) {
        Promise<Integer> p = Promise.value(1);
        p = p.cache();
        promises.add(p.map(v -> v + 1));
        promises.add(p.map(v -> v + 2));
    }

    final List<Integer> results = ExecHarness.yieldSingle(c ->
            ParallelBatch.of(promises).yield()
    ).getValueOrThrow();
}
```

When this test fails, the backtrace looks like this:

```
java.lang.NullPointerException
    at ratpack.exec.internal.CachingUpstream.yield(CachingUpstream.java:93)
    at ratpack.exec.internal.CachingUpstream.tryDrain(CachingUpstream.java:65)
    at ratpack.exec.internal.CachingUpstream.lambda$connect$0(CachingUpstream.java:116)
    at ratpack.exec.internal.CachingUpstream$$Lambda$58/1438461739.connect(Unknown Source)
    at ratpack.exec.internal.DefaultExecution.lambda$null$2(DefaultExecution.java:122)
    at ratpack.exec.internal.DefaultExecution$$Lambda$33/2092087501.execute(Unknown Source)
    at ratpack.exec.internal.DefaultExecution$SingleEventExecStream.exec(DefaultExecution.java:489)
    at ratpack.exec.internal.DefaultExecution.exec(DefaultExecution.java:216)
    at ratpack.exec.internal.DefaultExecution.exec(DefaultExecution.java:209)
    at ratpack.exec.internal.DefaultExecution.drain(DefaultExecution.java:179)
    at ratpack.exec.internal.DefaultExecution.<init>(DefaultExecution.java:92)
    at ratpack.exec.internal.DefaultExecController$1.lambda$start$0(DefaultExecController.java:195)
    at ratpack.exec.internal.DefaultExecController$1$$Lambda$7/1411892748.call(Unknown Source)
[...]

PR #1453, opened 2019-05-27
Add regression test for #1369

wololock:
Work in progress. Added regression test for #1369 The problem still exists, and from time to time one (or multiple) test hangs at releasing the countdown latch.

```
ratpack.exec.PromiseCachingSpec > can cache multiple subscribers (42) FAILED
    Method timed out after 10,00 seconds
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:997)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1304)
        at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:231)
        at ratpack.test.exec.internal.DefaultExecHarness.yield(DefaultExecHarness.java:61)
        at ratpack.test.exec.ExecHarness.yield(ExecHarness.java:122)
        at ratpack.test.exec.ExecHarness.yieldSingle(ExecHarness.java:149)
        at ratpack.exec.PromiseCachingSpec.can cache multiple subscribers(PromiseCachingSpec.groovy:245)

ratpack.exec.PromiseCachingSpec > can cache multiple subscribers (43) FAILED
    Method timed out after 10,00 seconds
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedInterruptibly(AbstractQueuedSynchronizer.java:997)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireSharedInterruptibly(AbstractQueuedSynchronizer.java:1304)
        at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:231)
        at ratpack.test.exec.internal.DefaultExecHarness.yield(DefaultExecHarness.java:61)
        at ratpack.test.exec.ExecHarness.yield(ExecHarness.java:122)
        at ratpack.test.exec.ExecHarness.yieldSingle(ExecHarness.java:149)
        at ratpack.exec.PromiseCachingSpec.can cache multiple subscribers(PromiseCachingSpec.groovy:245)
```

You can run the test with 

```
$ ./gradlew :ratpack-exec:clean :ratpack-exec:test --tests ratpack.exec.PromiseCachingSpec
```

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1453)

### DIFF
--- a/ratpack-exec/src/main/java/ratpack/exec/internal/CachingUpstream.java
+++ b/ratpack-exec/src/main/java/ratpack/exec/internal/CachingUpstream.java
@@ -136,7 +136,6 @@ public class CachingUpstream<T> implements Upstream<T> {
     Instant expiresAt;
     if (ttl.isNegative()) {
       expiresAt = null; // eternal
-      upstream = null; // release
     } else if (ttl.isZero()) {
       expiresAt = clock.instant().minus(Duration.ofSeconds(1));
     } else {
@@ -149,6 +148,10 @@ public class CachingUpstream<T> implements Upstream<T> {
     downstream.accept(result);
 
     tryDrain();
+
+    if (expiresAt == null) {
+      upstream = null; // release
+    }
   }
 
   private static class Cached<T> {
--- a/ratpack-exec/src/test/groovy/ratpack/exec/PromiseCachingSpec.groovy
+++ b/ratpack-exec/src/test/groovy/ratpack/exec/PromiseCachingSpec.groovy
@@ -21,7 +21,10 @@ import ratpack.exec.util.SerialBatch
 import ratpack.func.Action
 import ratpack.test.exec.ExecHarness
 import spock.lang.AutoCleanup
+import spock.lang.Issue
 import spock.lang.Specification
+import spock.lang.Timeout
+import spock.lang.Unroll
 
 import java.time.Duration
 import java.util.concurrent.ConcurrentLinkedQueue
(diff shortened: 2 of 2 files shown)


wololock:
@johnrengelman Could you please take a look at it? I applied your suggestion from #1369

wololock:
I have found something that might be related. I was trying to track what causes some of the test runs hang in the following line: https://github.com/ratpack/ratpack/blob/cb601bdd4a59d054ad2c23beec3689fc98ac6c58/ratpack-test/src/main/java/ratpack/test/exec/internal/DefaultExecHarness.java#L78 

It happens 1 or 2 times per all 51 executions, and from time to time the problem does not appear at all (very rarely though). So I started adding some println statements just to see which parts of the flow gets executed. I have found that if I add some stupid `System.out.println("anything")` before and after those two lines:

https://github.com/ratpack/ratpack/blob/6ddb4021261fb97b7d77122c914a9640dbd0e440/ratpack-exec/src/main/java/ratpack/exec/internal/CachingUpstream.java#L115-L116

the test stops freezing and it stops throwing NPE. I've run this test about 30 times in a row, not a single error.

```java
  @Override
  public void connect(Downstream<? super T> downstream) throws Exception {
    Cached<? extends T> cached = this.ref.get();
    if (needsFetch(cached)) {
      Promise.<T>async(d -> {
        System.out.println("Fetching async...");
        waiting.add(d);
        tryDrain();
        System.out.println("Fetching async done...");
      }).result(downstream::accept);
    } else {
      downstream.accept(cached.result);
    }
  }
```

I'm guessing that those two print statements extend the promise execution by a few nanoseconds - enough time to prevent the race condition from happening. 

I also wonder - if there is any specific reason that `CachedUpstream<T>.connect()` delegates fetching to an async Promise? I tried running the test case without this async Promise and there were no race conditions. I don't know if this is the right direction, but maybe this could solve the problem? Or does it cause problems in different areas? (All tests passed without `Promise.async` in this part.) What do you think?

PR #1459, opened 2019-06-28
WIP: Promise Zipping

johnrengelman:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1459)

### DIFF
--- a/ratpack-exec/src/main/java/ratpack/exec/util/PromiseZipper.java
+++ b/ratpack-exec/src/main/java/ratpack/exec/util/PromiseZipper.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ratpack.exec.util;
+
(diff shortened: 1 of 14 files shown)


drmaas (commented):
If the desired direction is to allow an execution initialization that applies to all promise via `ExecInitialier`, can you add a comment about that?

The other option would be to explicitly add it to the api like `ParallelBatch`.

johnrengelman:
I'm going to move this API addition to the next release (likely 2.0.0) as I don't think we have enough insight into how it will be used which could greatly  impact how we build out this API (static methods as above versus a new type similar to `ParallelBatch`). We shouldn't hold up the 1.7.0 release for this.

michaelschlies:
@johnrengelman Are you still thinking 1.8 for this? Is there anything in it that you could use some extra cycles moving dirt with?

drmaas (commented):
It would be great if there was a way to apply an `execInit` function for each forked execution, similar to how `ParallelBatch` does it.

ratpack-exec/src/main/java/ratpack/exec/util/PromiseZipper.java
```diff
@@ -0,0 +1,94 @@
...
+
+import ratpack.exec.ExecResult;
+import ratpack.exec.Promise;
+import ratpack.exec.util.internal.DefaultPromiseZipper;
+import ratpack.func.*;
+
+public interface PromiseZipper {
+
+
+  <T1, T2, R> Promise<R> yield(BiFunction<T1, T2, R> function);
```
drmaas: The `yield` and `yieldAll` could move into a `PromiseZipper2` interface to keep things consistent. Then `PromiseZipper` would be focused on `zip`.

Also, each `PromiseZipper*` interface could inherit another interface that provides an `execInit` function. Then `DefaultPromiseZipper` would use it on each forked execution.

Something like:
```
PromiseZipper zipper = Promises.zip(p1, p2, p3);
zipper.execInit(e -> e.add(Execution.current().get(String.class));
Promise<String> result = zipper.yield((r1, r2, r3) -> r1.val + r2.val);
```

PR #1460, opened 2019-06-29
Netty 4.1.37 update

johnrengelman:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1460)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/InstrumentedFixedChannelPoolHandler.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/InstrumentedFixedChannelPoolHandler.java
@@ -33,12 +33,6 @@ public class InstrumentedFixedChannelPoolHandler extends NoopFixedChannelPoolHan
   }
 
   @Override
-  public void channelCreated(Channel ch) throws Exception {
-    super.channelCreated(ch);
-    activeConnectionCount.increment();
-  }
-
-  @Override
   public void channelReleased(Channel ch) throws Exception {
     super.channelReleased(ch);
     activeConnectionCount.decrement();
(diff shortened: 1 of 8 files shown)


ratpack-core/src/main/java/ratpack/http/client/internal/InstrumentedFixedChannelPoolHandler.java
```diff
@@ -32,12 +32,6 @@ public InstrumentedFixedChannelPoolHandler(HttpChannelKey channelKey, int poolSi
     this.maxConnectionCount = poolSize;
   }
 
-  @Override
```
drmaas: Is this removed because `channelAcquired` is also called when a channel is created?

ratpack-core/src/main/java/ratpack/http/client/internal/InstrumentedSimpleChannelPoolHandler.java
```diff
@@ -29,12 +29,6 @@ public InstrumentedSimpleChannelPoolHandler(HttpChannelKey channelKey) {
     this.activeConnectionCount = new LongAdder();
   }
 
-  @Override
-  public void channelCreated(Channel ch) throws Exception {
```
drmaas: Same

ratpack-core/src/main/java/ratpack/http/client/internal/InstrumentedFixedChannelPoolHandler.java
johnrengelman: Correct. It was only overridden to record a metric. https://github.com/netty/netty/pull/9093

PR #1461, opened 2019-06-29
Groovy & Spock Update and new test fixtures for Remote APIs

johnrengelman:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1461)


ratpack-test/src/main/java/ratpack/test/mock/MockApi.java
```diff
@@ -0,0 +1,145 @@
...
+ *     } >> groovyHandler {
+ *       render("remote ok")
+ *     }
+ *     resp.body.text == "remote ok"
+ *   }
+ * }
+ * }</pre>
+ * @since 1.7.0
+ */
+public class MockApi implements EmbeddedApp {
```
drmaas: This is cool - can you add documentation and/or convenience method to mock out a chain of requests? We often have multitple endpoints to mock together. Something like:

```
HandlerFactory factory = request -> Handlers.chain(request, chain -> chain.get()); 
MockApi mockApi = MockApi.of(request);
```

Maybe something like:
```
public static HandlerFactory of(Action<? super Chain> action) {
    return request -> Handlers.chain(request, action);
}
```

johnrengelman: well, the intent here is to create a test fixture that you can combine both spock mock invocation and end-to-end http requests, with this you can excercise a full AUT talking to embedded apps over real network, but easily configure how the embedded apps handle each request inline to a spec.

Its not intended as a replacement for wiremock as that does offer some lower level controls for testing things like connection resets and such, but as a mocking response behavior, this maybe be a little cleaner.

johnrengelman: I don't see how providing a chain would work, in that case why not just make an embedded app?

drmaas: @johnrengelman makes sense - this feature will functionally doing similar things that WireMock also does, via a different and more lightweight mechanism. Perhaps calling out when to use this vs. something heavier like WireMock or BetaMax would be useful to end users.

PR #1466, opened 2019-07-19
ability to specify a predicate for retries

drmaas:
Use Case

This will allow users to retry only upon certain exceptions and conditions. For example, to retry only on connection timeout/reset errors, the predicate could match `IOException`.

TODO

~~* Add tests for predicate~~

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1466)

### DIFF
--- a/ratpack-exec/src/main/java/ratpack/exec/Promise.java
+++ b/ratpack-exec/src/main/java/ratpack/exec/Promise.java
@@ -2443,7 +2443,72 @@ public interface Promise<T> {
    * @since 1.7
    */
   default Promise<T> retry(RetryPolicy retryPolicy, BiAction<? super Integer, ? super Throwable> onError) {
-    return transform(up -> down -> DefaultPromise.retry(retryPolicy, up, down, onError));
+    return retryIf(Predicate.alwaysTrue(), retryPolicy, onError);
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


johnrengelman:
The `onError` function passed to `retry(RetryPolicy, Function)` (https://github.com/ratpack/ratpack/blob/master/ratpack-exec/src/main/java/ratpack/exec/Promise.java#L2445) allows for some level of control here. `onError` receives the attempt number and the underlying exception. Any exception thrown from `onError` bypasses the retry policy and is sent downstream.

drmaas:
Hm, that would also work, but seems a bit awkward and non-obvious.

drmaas:
@johnrengelman another option might be to overload retry - 

1) `retry(RetryPolicy retryPolicy, Predicate<? super Throwable> matcher)`
or 
2) `retryIf(Predicate<? super Throwable> matcher, RetryPolicy retryPolicy, BiAction<? super Integer, ? super Throwable> onError)`

johnrengelman:
I think I see the `retryIf(Predicate, ....)` pattern being the most appropriate implementation here and it should be fairly trivial to implement based on the semantics of `onError` that I mentioned above.

drmaas:
Updated to add a new method `retryIf`

PR #1467, opened 2019-07-19
Enable heartbeats on idle connection pool connections

drmaas:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1467)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/client/HttpClient.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/HttpClient.java
@@ -137,6 +137,15 @@ public interface HttpClient extends AutoCloseable {
   Duration getIdleTimeout();
 
   /**
+   * The action to execute on connection idle timeout.
+   * <p>
+   * If not set, the default action is to close the connection, which removes the connection from the pool.
+   *
+   * @since 1.8
+   */
+  IdleTimeoutAction getIdleTimeoutAction();
+
+  /**
    * The default read timeout value.
    *
    * @since 1.4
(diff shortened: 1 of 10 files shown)


johnrengelman:
So, when idle this checks the connection before releasing it? The idea being, that if the connection is still alive, then we can keep it around, even though it's been idle, and thus decrease latency if there is a sudden spike in requested connections? (Am I understanding this correctly?)

If my understanding is correct, then I think we should call this something different, like 'idleTestBeforeRelease' (that's a mouthful too). Maybe just `testOnIdle` is sufficient?

drmaas:
The idea here is to allow something other than closing the connection when idle timeout fires. So yes, it heartbeats the connection to keep it alive. These connections are *not* in use (a connection in use will reset the idle timer) and this is a way to prevent mass connection expiration.
Thinking on this some more, I think it would be better to provide the user with what action(s) they want to take on idle timeout. Options would be close (default), heartbeat, or potentially some custom action.

drmaas:
Another option is to expose the https://github.com/netty/netty/blob/netty-4.1.37.Final/transport/src/main/java/io/netty/channel/pool/ChannelHealthChecker.java api directly.

johnrengelman:
I like the idle of pluggable action for this.
Then we can provide defaults for `CLOSE` and `HEARTBEAT`. This action should be definable on 
`HttpClientSpec`

johnrengelman:
@drmaas with the change to `idleTimeout` to only  apply to connections that are not in-use (e.g. sitting in the connection pool) is this still applicable?
There's 2 ways to see this:
1. the timeout fires for a connection that has remained in the pool too long and we want to take an action.
2. An in-use connection has been sitting idle waiting for a server response (same as a read timeout) and we want to take an action.

Which problem are we trying to solve?

johnrengelman:
@drmaas is this still something you are seeing as a need?

drmaas:
I don't think so.

PR #1469, opened 2019-08-03
Fix HTTP Client with encoded query parameters

timyates:
We upgraded to 1.7.1, but discovered an issue whereby encoded query params sent back as part of a redirect Location header would be double encoded when the redirect was performed.

This should fix it (and doesn't break the fix introduced by https://github.com/ratpack/ratpack/issues/1436) for relative params.

Both relative and absolute redirect Locations were affected, and this should fix both 😎 

Not sure if there's a chance of an expedited 1.7.2 to include this?

Cheers, and keep up the good work!

Tim

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1469)


johnrengelman:
Thanks @timyates. I’ll try and get a 1.7.2 out this weekend.

timyates:
Awesome! Thanks @johnrengelman 😎👍

PR #1470, opened 2019-08-06
fix http client idle timeout to apply only to idleness in the pool

johnrengelman:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1470)

### DIFF
--- a/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientIdleTimeoutSpec.groovy
+++ b/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientIdleTimeoutSpec.groovy
@@ -16,6 +16,7 @@
 
 package ratpack.http.client
 
+import io.netty.channel.Channel
 import ratpack.error.ServerErrorHandler
 import ratpack.handling.Context
 import spock.lang.Unroll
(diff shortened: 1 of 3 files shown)


ratpack-core/src/test/groovy/ratpack/http/client/HttpClientIdleTimeoutSpec.groovy
```diff
@@ -116,6 +127,7 @@ class HttpClientIdleTimeoutSpec extends BaseHttpClientSpec {
 
     @Override
     void error(Context context, Throwable throwable) throws Exception {
+      throwable.printStackTrace()
```
drmaas: remove

johnrengelman: Whoops.

drmaas: @johnrengelman if you add another test where you sleep > timeout in between requests, you can assert that the channel is closed/null? and not in the pool.

johnrengelman: That's what the 2nd test case in `test client with idle timeout exceeded()`  tests. the `reused=false` condition on that test.

PR #1477, opened 2019-08-18
Replace referencecs to Ratpack's @Nullable with jetbrains

### LINKED ISSUE #1274: Annotate source with JSR305 (Nullable/Nonnull)
This would be helpful to Kotlin consumers of Ratpack to get better tooling assistance for when values may be null.

https://github.com/ratpack/ratpack/blob/master/ratpack-base/src/main/java/ratpack/api/Nullable.java exists but there is not an equivalent for Nonnull. `TypeQualifierNickname` or something similar can be used if internal annotations types are desired.

michaelschlies:
fixes #1274

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1477)


michaelschlies:
Replacing internal `@Nullable` could be breaking depending on if anyone is using reflection, this should be a 2.x change

ratpack-core/src/main/java/ratpack/file/FileSystemBinding.java
```diff
@@ -17,7 +17,7 @@
 package ratpack.file;
 
 import com.google.common.reflect.TypeToken;
-import ratpack.api.Nullable;
```
johnrengelman: @michaelschlies looks like your find/replace duplicated `;` throughout the code base.

michaelschlies: Good catch @johnrengelman, I'll clean those up this afternoon

michaelschlies: Ooops, finally got around to cleaning those up, CI running now

johnrengelman:
@michaelschlies looks like there is one failing spec with this change: `ratpack.guice.GuiceRegistrySpec.can inject null using ratpack's Nullable annotation`

johnrengelman:
I think you also need to rebase as there are some changes in master to fix test errors around some bad links.

michaelschlies:
Ah, looking into a @Nullable conflict with Guice, brb with this one.

PR #1478, opened 2019-08-18
Fix test order (latch countdown should happen after the desired effec…

michaelschlies:
…t of the ErrorListener)

@beckje01 @johnrengelman 
Fixed the spec execution
Correction from my previous comment in pr #1119

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1478)


michaelschlies:
Is there anything that holds this PR up since it is against the open branch Dan and Jeff have commits toward getting merged? - Thanks

PR #1484, opened 2019-09-20
Update links for HikariCP since github.io page is now unavailable

michaelschlies:
fixes #1483 in ratpack/master

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1484)

PR #1486, opened 2019-09-20
Remove sleep for 500ms and replace with additional latch.countDown()s

### LINKED ISSUE #1485: HierarchicalExecutionSpec tests intermittently fail under slow timing conditions in CircleCI
`HierarchicalExecutionSpec.current execution is parent during initializer of child` uses a sleep of 500ms to ensure ordering of parent-child relationships in execution forks. This will result in intermittently failing tests in CI but unlikely to present on local builds and test executions.

michaelschlies:
fixes #1485

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1486)


michaelschlies:
This will pass after #1484 is merged

PR #1489, opened 2019-10-04
Flush channel after writing PongWebSocketFrame

marcphilipp:
According to RFC 6455, section 5.5.2, an endpoint MUST send a Pong frame
in response to a Ping frame and it SHOULD respond as soon as is
practical. Currently, the Pong frames are queued until the channel is
flushed, e.g. by sending a `TextWebSocketFrame` via `WebSocket.send`.
This can lead to the queue growing indefinitely and the client closing
the connection because it assumes the server is not responding.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1489)


marcphilipp:
@ldaley Thanks for adding a test and merging this! 👍

PR #1492, opened 2019-10-24
[WIP] on read timeout, keep channels open when connection pooling

drmaas:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1492)


drmaas:
We found a condition where we end up closing a lot of connections in the pool unnecessarily when we encounter read timeouts. The connections in the pool are perfectly healthy, and we don't see a reason to close them. 

Instead, we would like these connections to remain open on read timeout, and allow them to be closed when they exceed the idle timeout or encounter an error per existing handlers.

drmaas:
I have been unable to get the test to pass - it seems something is still forcing the channel closed on read timeout.

drmaas:
Closing - I think the only reasonable thing to do on timeout is close the connection.

PR #1498, opened 2019-11-08
Http client proxy

johnrengelman:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1498)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/NoopFixedChannelPoolHandler.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/NoopFixedChannelPoolHandler.java
@@ -18,8 +18,11 @@ package ratpack.http.client.internal;
 
 import io.netty.channel.Channel;
 import io.netty.channel.pool.AbstractChannelPoolHandler;
+import io.netty.handler.proxy.HttpProxyHandler;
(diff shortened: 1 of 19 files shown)


ratpack-core/src/main/java/ratpack/http/client/internal/NoopFixedChannelPoolHandler.java
```diff
@@ -29,21 +32,26 @@
...
+      SocketAddress proxyAddress = new InetSocketAddress(proxy.getHost(), proxy.getPort());
+      ch.pipeline().addLast(new HttpProxyHandler(proxyAddress));
+    }
   }
 
   @Override
   public void channelReleased(Channel ch) throws Exception {
     if (ch.isOpen()) {
       ch.config().setAutoRead(true);
-      ch.pipeline().addLast(IdlingConnectionHandler.INSTANCE);
```
johnrengelman: @drmaas was this intentional or did one of us (probably me) keyboard slip here?

drmaas: @johnrengelman looks like it was you -  this is needed so that a channel released back into the pool gets closed if someone accidentally uses it.

PR #1500, opened 2019-11-25
Micrometer metrics module

jkschneider:
Adds a [Micrometer](micrometer.io) metrics module for dimensional metrics collection.

### DIFF
--- a/ratpack-micrometer-metrics/src/main/java/ratpack/micrometer/metrics/HandlerTags.java
+++ b/ratpack-micrometer-metrics/src/main/java/ratpack/micrometer/metrics/HandlerTags.java
@@ -1,0 +1,201 @@
+package ratpack.micrometer.metrics;
+
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import org.apache.commons.lang3.StringUtils;
+import ratpack.api.Nullable;
+import ratpack.handling.Context;
+import ratpack.http.Request;
+import ratpack.http.Response;
+import ratpack.http.Status;
+
+import java.util.function.BiFunction;
+import java.util.regex.Pattern;
+
+public final class HandlerTags {
+  public static final BiFunction<Request, Throwable, Tags> RECOMMENDED_BLOCKING_EXEC = (request, exception) -> Tags.of(
+    HandlerTags.method(request), HandlerTags.uri(request, null)
+  );
+
+  public static final BiFunction<Context, Throwable, Tags> RECOMMENDED_HANDLER = (context, exception) -> Tags.of(
+    HandlerTags.method(context.getRequest()), HandlerTags.uri(context.getRequest(), context.getResponse()),
+    HandlerTags.exception(exception), HandlerTags.status(context.getResponse()), HandlerTags.outcome(context.getResponse()));
+
+  private static final Tag URI_NOT_FOUND = Tag.of("uri", "NOT_FOUND");
+
+  private static final Tag URI_REDIRECTION = Tag.of("uri", "REDIRECTION");
+
+  private static final Tag URI_ROOT = Tag.of("uri", "root");
+
+  private static final Tag URI_UNKNOWN = Tag.of("uri", "UNKNOWN");
+
+  private static final Tag EXCEPTION_NONE = Tag.of("exception", "None");
+
+  private static final Tag STATUS_UNKNOWN = Tag.of("status", "UNKNOWN");
+
+  private static final Tag METHOD_UNKNOWN = Tag.of("method", "UNKNOWN");
+
+  private static final Pattern LEADING_SLASH_PATTERN = Pattern.compile("^/");
+
+  private static final Pattern TRAILING_SLASH_PATTERN = Pattern.compile("/$");
+
+  private static final Pattern MULTIPLE_SLASH_PATTERN = Pattern.compile("//+");
+
+  private HandlerTags() {
+  }
+
+  /**
+   * Creates a {@code method} tag based on the {@link Request#getMethod()
+   * method} of the given {@code request}.
+   * @param request the request
+   * @return the method tag whose value is a capitalized method (e.g. GET).
+   */
+  public static Tag method(Request request) {
+    return (request != null) ? Tag.of("method", request.getMethod().getName()) : METHOD_UNKNOWN;
+  }
+
+  /**
+   * Creates a {@code status} tag based on the status of the given {@code response}.
+   * @param response the HTTP response
+   * @return the status tag derived from the status of the response
+   */
+  public static Tag status(Response response) {
+    return (response != null) ? Tag.of("status", Integer.toString(response.getStatus().getCode())) : STATUS_UNKNOWN;
+  }
+
+  /**
+   * Creates a {@code uri} tag based on the URI of the given {@code request}. Falling back to {@code REDIRECTION} for 3xx responses,
+   * {@code NOT_FOUND} for 404 responses, {@code root} for requests with no path info, and {@code UNKNOWN} for all other requests.
+   * @param request the request
+   * @param response the response
+   * @return the uri tag derived from the request
+   */
+  public static Tag uri(Request request, @Nullable Response response) {
+    if (request != null) {
+      if (response != null) {
+        Status status = response.getStatus();
+        if (status != null) {
+          if (status.is3xx()) {
+            return URI_REDIRECTION;
+          }
+          if (status == Status.NOT_FOUND) {
+            return URI_NOT_FOUND;
+          }
+        }
+      }
+      String pathInfo = getPathInfo(request);
+      if (pathInfo.isEmpty()) {
+        return URI_ROOT;
+      } else {
+        return Tag.of("uri", pathInfo);
+      }
+    }
+    return URI_UNKNOWN;
+  }
+
+  private static String getPathInfo(Request request) {
+    String pathInfo = request.getPath();
+    String uri = StringUtils.isNotBlank(pathInfo) ? pathInfo : "/";
+    uri = MULTIPLE_SLASH_PATTERN.matcher(uri).replaceAll("/");
+    uri = LEADING_SLASH_PATTERN.matcher(uri).replaceAll("");
+    return TRAILING_SLASH_PATTERN.matcher(uri).replaceAll("");
+  }
+
+  /**
+   * Creates a {@code exception} tag based on the {@link Class#getSimpleName() simple
+   * name} of the class of the given {@code exception}.
+   * @param exception the exception, may be {@code null}
+   * @return the exception tag derived from the exception
+   */
+  public static Tag exception(Throwable exception) {
+    if (exception != null) {
+      String simpleName = exception.getClass().getSimpleName();
+      return Tag.of("exception", StringUtils.isNotBlank(simpleName) ? simpleName : exception.getClass().getName());
+    }
+    return EXCEPTION_NONE;
+  }
+
+  /**
+   * Creates an {@code outcome} tag based on the status of the given {@code response}.
+   * @param response the HTTP response
+   * @return the outcome tag derived from the status of the response
+   */
+  public static Tag outcome(Response response) {
+    Outcome outcome = (response != null) ? Outcome.forStatus(response.getStatus().getCode()) : Outcome.UNKNOWN;
+    return outcome.asTag();
+  }
+
+  /**
+   * The outcome of an HTTP request.
+   */
+  public enum Outcome {
+
+    /**
+     * Outcome of the request was informational.
+     */
+    INFORMATIONAL,
+
+    /**
+     * Outcome of the request was success.
+     */
+    SUCCESS,
+
+    /**
+     * Outcome of the request was redirection.
+     */
+    REDIRECTION,
+
+    /**
+     * Outcome of the request was client error.
+     */
+    CLIENT_ERROR,
+
+    /**
+     * Outcome of the request was server error.
+     */
+    SERVER_ERROR,
+
+    /**
+     * Outcome of the request was unknown.
+     */
+    UNKNOWN;
+
+    private final Tag tag;
+
+    Outcome() {
+      this.tag = Tag.of("outcome", name());
+    }
+
+    /**
+     * Returns the {@code Outcome} as a {@link Tag} named {@code outcome}.
+     * @return the {@code outcome} {@code Tag}
+     */
+    public Tag asTag() {
+      return this.tag;
+    }
+
+    /**
+     * Return the {@code Outcome} for the given HTTP {@code status} code.
+     * @param status the HTTP status code
+     * @return the matching Outcome
+     */
+    public static Outcome forStatus(int status) {
+      if (status >= 100 && status < 200) {
+        return INFORMATIONAL;
+      }
+      else if (status >= 200 && status < 300) {
+        return SUCCESS;
+      }
+      else if (status >= 300 && status < 400) {
+        return REDIRECTION;
+      }
+      else if (status >= 400 && status < 500) {
+        return CLIENT_ERROR;
+      }
+      else if (status >= 500 && status < 600) {
+        return SERVER_ERROR;
+      }
+      return UNKNOWN;
+    }
+  }
+}
--- a/ratpack-micrometer-metrics/src/main/java/ratpack/micrometer/metrics/internal/DefaultBlockingExecTimingInterceptor.java
+++ b/ratpack-micrometer-metrics/src/main/java/ratpack/micrometer/metrics/internal/DefaultBlockingExecTimingInterceptor.java
@@ -1,0 +1,44 @@
+package ratpack.micrometer.metrics.internal;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Timer;
+import ratpack.exec.ExecInterceptor;
+import ratpack.exec.Execution;
+import ratpack.func.Block;
+import ratpack.http.Request;
+import ratpack.micrometer.metrics.BlockingExecTimingInterceptor;
+import ratpack.micrometer.metrics.MicrometerMetricsConfig;
+
+import java.util.Optional;
+
+public class DefaultBlockingExecTimingInterceptor implements BlockingExecTimingInterceptor {
+  private final MicrometerMetricsConfig config;
+  private final MeterRegistry meterRegistry;
+
+  DefaultBlockingExecTimingInterceptor(MicrometerMetricsConfig config, MeterRegistry meterRegistry) {
+    this.config = config;
+    this.meterRegistry = meterRegistry;
+  }
+
+  @Override
+  public void intercept(Execution execution, ExecInterceptor.ExecType type, Block executionSegment) throws Exception {
+    if (type == ExecInterceptor.ExecType.BLOCKING) {
+      Optional<Request> requestOpt = execution.maybeGet(Request.class);
+      if (requestOpt.isPresent()) {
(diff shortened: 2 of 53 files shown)


johnrengelman:
This is....awesome. Thanks!
Looks like there is a dependency  conflict here:
```
Dependency resolution failed because of conflicts between the following modules:
   - org.hdrhistogram:HdrHistogram:2.1.9
   - org.hdrhistogram:HdrHistogram:2.1.11

org.hdrhistogram:HdrHistogram:2.1.11
   variant "runtime" [
      org.gradle.status             = release (not requested)
      org.gradle.usage              = java-runtime (not requested)
      org.gradle.component.category = library (not requested)
   ]
   Selection reasons:
      - By conflict resolution : between versions 2.1.9 and 2.1.11

org.hdrhistogram:HdrHistogram:2.1.11
\--- io.micrometer:micrometer-core:1.3.1
     \--- project :ratpack-micrometer-metrics
          \--- testCompile

org.hdrhistogram:HdrHistogram:2.1.9 -> 2.1.11
\--- com.netflix.hystrix:hystrix-core:1.5.13
     \--- project :ratpack-hystrix
          \--- testCompile
```

ratpack-micrometer-metrics/src/main/java/ratpack/micrometer/metrics/HandlerTags.java
```diff
@@ -0,0 +1,201 @@
...
+          if (status == Status.NOT_FOUND) {
+            return URI_NOT_FOUND;
+          }
+        }
+      }
+      String pathInfo = getPathInfo(request);
+      if (pathInfo.isEmpty()) {
+        return URI_ROOT;
+      } else {
+        return Tag.of("uri", pathInfo);
```
smcadams86: won't this generate high cardinality metrics when path variables are used in requests?

drmaas: Agreed - we need a way to either 1) generate tags based on path tokens or 2) make this configurable. @smcadams86 can you share the code you have that uses path tokens instead of resolved values?

drmaas:
This looks great - I'd like to see a server interceptor and client interceptor (which could be configurable and be able to be enabled/disabled) added to automatically be able to capture server request and client request metrics.

ratpack-micrometer-metrics/src/main/java/ratpack/micrometer/metrics/HandlerTags.java
smcadams86: I grabbed the `description` from the `pathBinding` object: https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/main/java/ratpack/path/PathBinding.java#L88

jkschneider: Yes, thank you @smcadams86! Meant to get back to this case. Appreciate the code sample as well.

jkschneider:
@drmaas My knowledge of ratpack is limited. If I understand correctly, is server interceptor is handled by `RequestTimingHandler` (or is there a separate thing?). Do you imagine a client-side metrics user experience similar to [MDCInterceptor](https://ratpack.io/manual/current/api/index.html?ratpack/logging/MDCInterceptor.html)?

jkschneider:
@johnrengelman I suppose I'll add the `dependencyRule` for `HdrHistogram` because if someone is currently using `ratpack-hystrix` and doesn't include this essentially optional module for Micrometer, it could be breaking if they are using `RollingCollapserBatchSizeDistributionStream`.

jkschneider:
There may be a little issue with the way Ratpack deals with path bindings. Consider the handler spec:

```
handlers { 
  prefix('foo') {
    path('baz/:id') { ctx ->
      render '...'
    }
  }
}
```

```
get('/foo/baz/123')  // correctly hits the path binding
get('///foo/baz/123') // does not
```

If this behavior is ever changed, there _won't_ be a required change to this new module, because it is written to attempt to eagerly strip extra leading and trailing slashes.

ratpack-micrometer-metrics/src/main/java/ratpack/micrometer/metrics/HandlerTags.java
jkschneider: Resolved in 7e0b36c.

ratpack-micrometer-metrics/src/main/java/ratpack/micrometer/metrics/internal/DefaultBlockingExecTimingInterceptor.java
```diff
@@ -0,0 +1,44 @@
...
+
+  DefaultBlockingExecTimingInterceptor(MicrometerMetricsConfig config, MeterRegistry meterRegistry) {
+    this.config = config;
+    this.meterRegistry = meterRegistry;
+  }
+
+  @Override
+  public void intercept(Execution execution, ExecInterceptor.ExecType type, Block executionSegment) throws Exception {
+    if (type == ExecInterceptor.ExecType.BLOCKING) {
+      Optional<Request> requestOpt = execution.maybeGet(Request.class);
```
johnrengelman: There's a comment that you added above for that says that you don't have access to the `Context`, but you should be able to just change this to `execution.maybeGet(Context.class)`. The `Context` will then contain both the `Request` and `Response`.

ratpack-micrometer-metrics/src/main/java/ratpack/micrometer/metrics/internal/DefaultBlockingExecTimingInterceptor.java
```diff
@@ -0,0 +1,44 @@
...
+      if (requestOpt.isPresent()) {
+        Request request = requestOpt.get();
+        Timer.Sample sample = Timer.start(meterRegistry);
+        try {
+          executionSegment.execute();
+          sample.stop(meterRegistry.timer("http.blocking.execution",
```
johnrengelman: Side comment - one of the biggest things we are missing is a central place that lists all these measurement names and what they represent (docs perhaps). I also wonder if we should make this pluggable in some way (though that could make this overly-complex). 
Are there any standards out there metrics naming? (like CNCF or anything?)

jkschneider: By pluggable do you mean that the user can configure the names to be something else? If so, this is already possible with `MeterFilter`. This interface was built with the idea that users could change names, add/remove/change tags, and assign different base units and description text. There's a config item in `MicrometerMertricsConfig` here that allows Ratpack users to add `MeterFilters`.

There is no open standard for metrics naming. Some notes however:
1. Micrometer uses dot separated names by convention, and each registry implementation is packed with a name normalizer that transforms its style to something the monitoring system expects (e.g. `http.server.requests` would appear as a set of timeseries prefixed with `http_server_requests_seconds` in Prometheus). The idioms of each monitoring system are mutually exclusive, so it's best to follow the dot convention in frameworks and let Micrometer do the dirty work.
2. When possible, this PR follows Spring Boot metrics names, which were in turn copied in Micronaut, etc. So a sort of informal standard is in play already.

ratpack-micrometer-metrics/src/main/java/ratpack/micrometer/metrics/internal/DefaultBlockingExecTimingInterceptor.java
jkschneider: Whaaaaaatt... This changes everything @johnrengelman. Removed all the duplication around setting a different tag handler for blocking exec vs. execution listeners now, so `http.server.requests` and `http.blocking.execution` will align on tags (which is better anyway). Fixed in c535c7ce8d8d973d694d0f70523a4d9ac1cd1f4f.

drmaas:
@jkschneider there are a lot (too many?) different types of interceptors - aop, execution, client, and server interceptors.

Clients have request and response interceptors which can be used to collect metrics automatically for paths. They are just of type `Action<? super RequestSpec>` and `Action<? super HttpResponse>`. See https://github.com/ratpack/ratpack/blob/master/ratpack-core/src/test/groovy/ratpack/http/client/HttpClientInterceptorSpec.groovy#L41

I misspoke about server requests. Servers can do the same, but not with interceptors - just a regular handler that gets called for each request and delegates to the next in the chain -  as you said, like `RequestTimingHandler`.

jkschneider:
> I misspoke about server requests. Servers can do the same, but not with interceptors - just a regular handler that gets called for each request and delegates to the next in the chain - as you said, like RequestTimingHandler.

@drmaas It would be nice to be able to figure out how to time 404s. AFAICT, because Ratpack has a special internal interceptor guaranteed to be last in the interceptor chain, I never get to observe a 404 from the perspective of instrumentation.

> Clients have request and response interceptors which can be used to collect metrics automatically for paths.

OK, will see how well I can understand this. If it gets too inscrutable, maybe we could address client instrumentation in a separate PR so it can be reviewed separately.

johnrengelman:
> because if somebody hits a bad path with lots of variety, e.g. /api/doesnotexist/1, /api/doesnotexist/2, etc. could blow up tag cardinality.

I think in general we want to use the `PathBinding` for the tag instead of the actual path to reduce this. In the case of the default handlers, I believe the `PathBinding` with be `null` since there was no match within the chain.

johnrengelman:
> Ratpack has a special internal interceptor guaranteed to be last in the interceptor chain, I never get to observe a 404 from the perspective of instrumentation

I'm not sure what this is saying, but I'll try to clarify what I think it is saying. The ratpack default handlers are not _special_ in any way. They are simply appended to the end of the chain. As for observing, placing an `all` handler at the front of the chain and then using the `Context.onClose` method, the code will be invoked, regardless of if it was handled by  a user provided handler or a built-in one.

drmaas:
@jkschneider a few thoughts if you are interested - I can post it here or in Slack.

* I can share sample client interceptor metrics code
* I can share a sample of using `onClose` to time metrics
* I can share a `PathBinding` example

drmaas:
Interceptor code samples are in Kotlin. I was not able to determine how to determine variable path elements in client requests - internally we scrub a list of known path patterns.

```
    private val regex1 = Regex(pattern = "/a/.*")
    private val regex2 = Regex(pattern = "/b/.*/get_details")
    private val regex3 = Regex(pattern = "/c/.*/defaults")

    private fun scrubPath(path: String): String {
        return path
            .replace(regex1, replacement = "/a/v1/{id}")
            .replace(regex2, replacement = "/b/{id}/get_details")
            .replace(regex3, replacement = "/c/{id}/defaults")
    }
```
```
data class MetricContextHolder(
    val timer: Timer.Sample,
    val method: String,
    val path: String
)
```
```
    override fun request(spec: RequestSpec) {
        registrySupplier.get()
            .ifPresent { execution ->
                val timer = Timer.start()
                val metricHolder = MetricContextHolder(
                    timer = timer,
                    method = spec.method.name,
                    path = scrubPath(spec.uri.path)
                )
                execution.add(metricHolder)
            }
    }
```
```
    override fun response(response: HttpResponse) {
        registrySupplier.get()
            .ifPresent { execution ->
                execution
                    .maybeGet(MetricContextHolder::class.java)
                    .ifPresent { metricContextHolder ->
                        metricContextHolder.timer.stop(
                            Timer.builder(HTTP_INTEGRATION_REQUESTS)
                                .publishPercentiles(0.50, 0.75, 0.95, 0.99, 0.999)
                                .publishPercentileHistogram()
                                .tags(
                                    listOf(
                                        ImmutableTag("status_code", response.statusCode.toString()),
                                        ImmutableTag("method", metricContextHolder.method),
                                        ImmutableTag("path", metricContextHolder.path)
                                    )
                                )
                                .register(compositeMeterRegistry)
                        )
                    }
            }
    }
```
```
    override fun error(e: Throwable) {
        registrySupplier.get()
            .ifPresent { execution ->
                execution
                    .maybeGet(MetricContextHolder::class.java)
                    .ifPresent { metricContextHolder ->
                        metricContextHolder.timer.stop(
                            Timer.builder(HTTP_INTEGRATION_EXCEPTIONS)
                                .publishPercentiles(0.50, 0.75, 0.95, 0.99, 0.999)
                                .publishPercentileHistogram()
                                .tags(
                                    listOf(
                                        ImmutableTag("exception", e.javaClass.simpleName),
                                        ImmutableTag("method", metricContextHolder.method),
                                        ImmutableTag("path", metricContextHolder.path)
                                    )
                                )
                                .register(compositeMeterRegistry)
                        )
                    }
            }
    }
```

drmaas:
Server side metrics:

```
@FunctionalInterface
public interface RequestTimingHandler extends Handler {
}
```
```
public interface MetricTagBuilder {
  String uuidRegex = "/([0-9a-fA-F]{8}-?[0-9a-fA-F]{4}-?[0-9a-fA-F]{4}-?[0-9a-fA-F]{4}-?[0-9a-fA-F]{12})";
  String idRegex = "/([0-9]+)";

  default String buildPathTag(MicrometerConfig config, String requestPath) {
    String shortenedPath = config.getPathPrefix().map(it -> requestPath.replaceAll(it, "")).orElse(requestPath);
    String tagName = shortenedPath.equals("") ? "/" : shortenedPath;

    if (config.getRequestMetricGroups() != null) {
      for (Map.Entry<String, String> metricGrouping : config.getRequestMetricGroups().entrySet()) {
        if (shortenedPath.matches(metricGrouping.getValue())) {
          tagName = metricGrouping.getKey();
          break;
        }
      }
    }

    return tagName
      .replaceAll(uuidRegex,"/{id}")
      .replaceAll(idRegex,"/{id}");
  }
}
```
```
public class MicrometerRequestTimingHandler implements RequestTimingHandler, MetricTagBuilder {

  private static final String HTTP_SERVER_REQUESTS = "http.server.requests";
  private final CompositeMeterRegistry meterRegistry;
  private final MicrometerConfig config;

  public MicrometerRequestTimingHandler(CompositeMeterRegistry meterRegistry, MicrometerConfig config) {
    this.meterRegistry = meterRegistry;
    this.config = config;
  }

  @Override
  public void handle(final Context context) {
    context.onClose(outcome -> {
      final String method = outcome.getRequest().getMethod().getName();
      final String path = buildPathTag(config, context.getPathBinding().getDescription());
      final String statusCode = String.valueOf(outcome.getResponse().getStatus().getCode());
      final String statusCodeGroup = Metrics.RESPONSES_PREFIX + statusCode.substring(0, 1) + "xx";

      Tags tags = Tags.of(
          Metrics.PATH, path,
          Metrics.METHOD, method,
          Metrics.STATUS_CODE, statusCode,
          Metrics.STATUS_CODE_GROUP, statusCodeGroup
        )
        .and(additionalTags(context));

      Timer.builder(HTTP_SERVER_REQUESTS)
        .publishPercentiles(0.50, 0.75, 0.95, 0.98, 0.99, 0.999)
        .publishPercentileHistogram()
        .tags(tags)
        .register(meterRegistry)
        .record(outcome.getDuration().toMillis(), TimeUnit.MILLISECONDS);

    });
    context.next();
  }

  protected Tags additionalTags(Context context) {
    return Tags.empty();
  }
}
```
Provider:
```
public class RequestTimingHandlerProvider implements Provider<RequestTimingHandler> {
  private final CompositeMeterRegistry meterRegistry;
  private final MicrometerConfig config;

  @Inject
  public RequestTimingHandlerProvider(CompositeMeterRegistry meterRegistry, MicrometerConfig config) {
    this.meterRegistry = meterRegistry;
    this.config = config;
  }

  @Override
  public RequestTimingHandler get() {
    return config.isRequestTimingMetrics() ? new MicrometerRequestTimingHandler(meterRegistry, config) : Context::next;
  }
}
```
Module:
```
    Provider<RequestTimingHandler> handlerProvider = getProvider(RequestTimingHandler.class);
    Multibinder.newSetBinder(binder(), HandlerDecorator.class).addBinding().toProvider(() -> HandlerDecorator.prepend(handlerProvider.get()));

```

jkschneider:
Returning to this to finish it up...

@johnrengelman The path bindings are non-null even for unroutable requests. When I examine the path bindings on the `Context` object for unroutable requests, there is just a single `RootPathBinding`, where routable requests always seem to have that plus one or more `DefaultPathBinding`. At any rate, no way to determine that this is the case without reflection. I also noticed that `context.getPathBinding().getBoundTo()` is empty hopefully _only_ when the request is ultimately unroutable? So that's the test I'm using currently to apply `uri=NOT_FOUND` tag.

jkschneider:
@drmaas I think I'm being a little dense here, I still am not quite sure how to propagate state between client requests and responses. I've tried this, but the header isn't visible on the response object, and I really don't think it's good to have to use a concurrent map like this, afraid contention on it will limit throughput. It's not clear to me from your example what `execution` is?

```java
@Override
  public void execute(HttpClientSpec httpClientSpec) {
    Map<String, Timer.Sample> samplesByRequest = new ConcurrentHashMap<>();
    Map<String, RequestSpec> specByRequest = new ConcurrentHashMap<>();

    httpClientSpec.requestIntercept(requestSpec -> {
      String requestId = UUID.randomUUID().toString();
      requestSpec.headers(h -> h.set(MICROMETER_REQUEST_ID, requestId));
      samplesByRequest.put(requestId, Timer.start(registry));
      specByRequest.put(requestId, requestSpec);
    });

    httpClientSpec.responseIntercept(responseSpec -> {
      String requestId = responseSpec.getHeaders().get(MICROMETER_REQUEST_ID);
      Timer.Sample sample = samplesByRequest.remove(requestId);
      RequestSpec request = specByRequest.remove(requestId);
      if (sample != null) {
        sample.stop(registry.timer("http.client.requests",
          config.getClientRequestTags().apply(request, responseSpec)));
      }
    });
  }
```

drmaas:
Hi @jkschneider - I will hopefully help to explain the issues you're seeing.

Definition of an `Execution` (from the docs): 
```A logical operation, such as servicing a request, that may be comprised of non contiguous units of work.
...
Each execution is a {@link MutableRegistry}.
 * Objects can be added to this registry and then later retrieved at any time during the execution.
 * The registry storage can be leveraged via an {@link ExecInterceptor} to manage thread local storage for execution segments.
``` 

I forgot to include what `registrySupplier` is. It is a class-level field.
```
private val registrySupplier: Supplier<Optional<Execution>> = Supplier { Execution.currentOpt() }
```

So the interceptors above simply fetch the current execution for this request. If it's not there for some weird reason - don't do anything. If it is, since the execution can hold state - use it to stash what you need.

Hope that helps!

jkschneider:
@drmaas 
> Hope that helps!

Tried working with currnet `Execution`, but the instance always appears different in request and response interceptors to me. So what I'm stuffing on the execution in the request handler isn't what's available in the response interceptor.

`HttpClientTimingActionSpec` fails as a result currently.

drmaas:
@jkschneider thanks for pointing that out - there is one additional component that we needed to put into place to allow an `Execution` (no matter if request/response are on same or different executions) to contain the state of the request. We had to add an `ExecInitializer`:

```
class MetricsPropagationExecInitializer : ExecInitializer {
    override fun init(execution: Execution) {
        execution.maybeParent()
            .flatMap { parent -> parent.maybeGet(MetricContextHolder::class.java) }
            .ifPresent { execution.add(it) }
    }
}
```

My bad for omitting this crucial component. Try bringing this in to see if it resolves your issue.

johnrengelman:
@jkschneider thanks for the effort with this, I'm curious if you could summarize what is left to to get this to a point where it can be considered to go in? I'm looking at cutting the 1.8 release soon, so trying to judge if this should slide to a 1.9 or if we can push it over the finish line.

PR #1501, opened 2019-11-27
Add support for binary websocket frames

### LINKED ISSUE #746: Support for BinaryWebSocketFrame
Hey,

Currently WebSocketEngine doesn't handle binary frames,

If possible, please add binary message support in WebSocketEngine#HandshakeFutureListener,

See issue #646 

Thanks

marcphilipp:
The `WebSocket` now has `sendText(String/ByteBuf)` and
`sendBinary(ByteBuf)` methods. The old `send(String/ByteBuf)` methods
are deprecated to make it clear that `sendText()` or `sendBinary()`
should be used instead.

The `WebSocketMessage` now has new `isBinary()` and `getContent()`
methods. The `getText()` method is still supported and decodes the
content into a utf8 `String`. The only drawback of this approach is
that existing implementations of `WebSocketHandler` will start
receiving binary frames in their `onMessage(WebSocketMessage)` methods.
An alternative would be to introduce a new `onBinaryMessage` method
with an empty default implementation, deprecate `onMessage`, and
delegate to `onMessage` from the default implementation of
`onTextMessage` for backwards compatibility.

Resolves #746.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1501)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/websocket/WebSocket.java
+++ b/ratpack-core/src/main/java/ratpack/websocket/WebSocket.java
@@ -29,10 +29,27 @@ public interface WebSocket {
 
   boolean isOpen();
 
+  /**
+   * @deprecated please use {@link #sendText(String)}
+   */
+  @Deprecated
   @NonBlocking
   void send(String text);
 
+  /**
+   * @deprecated please use {@link #sendText(ByteBuf)} or {@link #sendBinary(ByteBuf)}
+   */
+  @Deprecated
   @NonBlocking
   void send(ByteBuf text);
+
+  @NonBlocking
+  void sendText(String text);
+
+  @NonBlocking
+  void sendText(ByteBuf text);
+
+  @NonBlocking
+  void sendBinary(ByteBuf text);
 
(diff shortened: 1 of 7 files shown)


marcphilipp:
@ldaley Friendly ping. 🙂

johnrengelman:
Sorry, I'm going to be super dense on this....if this changes breaks existing user's code, then it will have to be pushed to Ratpack 2.0 unless we provide the backwards compatible implementation.

marcphilipp:
I'd say the chance that binary web socket frames were previously sent by clients but not handled by the server is rather slim. If you'd prefer, we could use two separate methods `onTextMessage` and `onBinaryMessage` (as I outlined in the description) instead. WDYT?

ldaley:
I'll add this next week.

marcphilipp:
Closing due to lack of progress. Let me know if I should reopen. 🙂

negram:
@johnrengelman, @Idaley can we resurrect this?

PR #1509, opened 2020-02-14
Don't send CONTENT_LENGTH if TRANSFER_ENCODING is CHUNKED

paplorinc:
According to https://tools.ietf.org/html/rfc7230#section-3.3.3:

> Screen Shot 2020-02-14 at 12 49 00 PM

This was needed to be able to upgrade Netty to > [4-1-44](https://netty.io/news/2019/12/18/4-1-44-Final.html)

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1509)

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -18,8 +18,8 @@ ext {
   commonVersions = [
     remote           : "0.7",
     slf4j            : "1.7.25",
-    netty            : "4.1.37.Final",
-    nettyTCNative    : "2.0.25.Final",
+    netty            : "4.1.45.Final",
+    nettyTCNative    : "2.0.28.Final",
     guava            : "21.0",
     groovy           : "2.5.4",
     pac4j            : "1.8.8",
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/RequestActionSupport.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/RequestActionSupport.java
@@ -16,6 +16,7 @@
 
 package ratpack.http.client.internal;
 
+import com.google.common.collect.Iterables;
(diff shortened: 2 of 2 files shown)


gradle/dependencies.gradle
```diff
@@ -18,8 +18,8 @@ ext {
   commonVersions = [
     remote           : "0.7",
     slf4j            : "1.7.25",
-    netty            : "4.1.37.Final",
-    nettyTCNative    : "2.0.25.Final",
+    netty            : "4.1.45.Final",
```
paplorinc: See: 
* https://netty.io/news/2019/12/18/4-1-44-Final.html
* https://netty.io/news/2020/01/13/4-1-45-Final.html

ratpack-core/src/main/java/ratpack/http/client/internal/RequestActionSupport.java
```diff
@@ -166,10 +167,17 @@ private void finalizeHeaders() {
     }
     int contentLength = requestConfig.body.readableBytes();
     if (contentLength > 0) {
-      requestConfig.headers.set(HttpHeaderConstants.CONTENT_LENGTH, Integer.toString(contentLength));
+      // See https://github.com/netty/netty/pull/9865/files#diff-bf7a179a319b29e0262d5badaf3770e0R634-R650
```
paplorinc: After the upgrade these tests were failing, the following changes made them pass again:

ldaley:
This isn't needed. We should be setting the content length here.

The problem was that the user code was explicitly setting `transfer-encoding: chunked`, but then not sending a chunked response. We could potentially deal with that better, but that's a separate issue.

PR #1513, opened 2020-03-31
Upgrade vulnerable dependencies

jdbevan:
So I was going to try and update a few of the other vulnerable dependencies but got hit with a load of dependency resolution conflicts in ratpack-consul which are beyond my capabilities to fix

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1513)

### DIFF
--- a/gradle/dependencies.gradle
+++ b/gradle/dependencies.gradle
@@ -18,7 +18,7 @@ ext {
   commonVersions = [
     remote           : "0.7",
     slf4j            : "1.7.25",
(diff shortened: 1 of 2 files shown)


johnrengelman:
Upgraded in https://github.com/ratpack/ratpack/commit/0e83316c89acd2e075ad00c6f93b5faa5553b446.
Not adding the `owasp` plugin because it causes issues in our build.

jdbevan:
Thanks @johnrengelman - any ETA on the next release?

PR #1514, opened 2020-04-14
Add Promise.onComplete.

### LINKED ISSUE #1464: Proposal: Promise onComplete actions
Proposing to add a method to `Promise` to allow taking action on a promise when it completes without a value.

```
  default Promise<T> onComplete(Block block) {
    return transform(up -> down ->
      up.connect(down.onComplete(() -> {
        try {
          block.execute();
        } catch (Throwable e) {
          down.error(e);
        }
      }))
    );
  }
```

This would be useful in a chain where a Promise from a dependent library issued on `onError`, thus terminating the promise, and the user code wants to take action when a promise is terminated without a value. There might be other scenarios as well where a promise would terminate, and a user would want to execute and arbitrary block of code when that happens.

johnrengelman:
Closes #1464

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1514)

### DIFF
--- a/ratpack-exec/src/main/java/ratpack/exec/Promise.java
+++ b/ratpack-exec/src/main/java/ratpack/exec/Promise.java
@@ -447,7 +447,7 @@ public interface Promise<T> {
   }
 
   /**
-   * Specifies the action to take if the an error occurs trying to produce the promised value.
+   * Specifies the action to take if an error occurs trying to produce the promised value.
    * <p>
    * If the given action throws an exception, the original exception will be rethrown with the exception thrown
    * by the action added to the suppressed exceptions list.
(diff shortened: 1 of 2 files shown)


johnrengelman:
@drmaas does this look right for you? Can you maybe share an example of how you envision this being useful in Resilience4j?

drmaas:
@johnrengelman It was mainly for scenarios where promises complete (perhaps due to `onError` or `route`) without a value or exception being present.

Examples:

https://github.com/ratpack/ratpack/blob/master/ratpack-exec/src/main/java/ratpack/exec/Promise.java#L396

https://ratpack.io/manual/current/api/ratpack/exec/Downstream.html#complete--

https://ratpack.io/manual/current/api/ratpack/exec/ExecResult.html#isComplete--

https://ratpack.io/manual/current/api/ratpack/exec/Promise.html#route-ratpack.func.Predicate-ratpack.func.Action-

PR #1554, opened 2020-07-02
Bring ExecController construction in public API and support custom execution type bindings

johnrengelman:


johnrengelman:
Implemented part of this here - https://github.com/ratpack/ratpack/pull/1632

PR #1560, opened 2020-07-06
Properly handling Expect: 100-continue when request streaming as proxy.

johnrengelman:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1560)


ldaley:
@johnrengelman can you please rebase this onto 1.9?

johnrengelman:
@ldaley done

ratpack-core/src/test/groovy/ratpack/http/client/HttpReverseProxySpec.groovy
```diff
@@ -420,8 +424,160 @@ connection: close
     status << noBodyResponseStatuses()
   }
 
+  def "can proxy a 100-continue request"() {
```
ldaley, line 427: Missing some coverage for redirect case IIUC. Would be worth also testing a double redirect case.

ldaley (changes requested):

johnrengelman:
@ldaley - Not sure what the appropriate behavior for a redirect would be....`100-Continue` is used for `PUT/POST` messages, but it's not really valid to redirect a user on those method types. In fact sending a 301/302 results in us changing the request method from `POST` to `GET` on the redirect.
I think `307 POST` is supposed to be used for redirecting a `POST` but the spec says that should require user acknowledgement.

ratpack-core/src/test/groovy/ratpack/http/client/HttpReverseProxySpec.groovy
johnrengelman: Added specs for this.

ldaley (approved):

PR #1579, opened 2021-01-26
Set checkstyle plugin config dir via configDirectory option

rpygithub:
Update references to the config directory in the Checkstyle configuration files to use [the provided](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.quality.Checkstyle.html) 'config_loc' variable accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1579)


ldaley:
Thank you @rpygithub

PR #1581, opened 2021-02-10
Use netty nio DNS resolver by default in HttpClient

johnrengelman:
Allows for overriding the implementation by adding AddressResolverGroup
to the Registry.

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1581)


ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
```diff
@@ -49,9 +51,13 @@
   private final HttpChannelPoolMap channelPoolMap = new HttpChannelPoolMap() {
     @Override
     protected ChannelPool newPool(HttpChannelKey key) {
+
       Bootstrap bootstrap = new Bootstrap()
         .remoteAddress(key.host, key.port)
         .group(key.execution.getEventLoop())
+        .resolver(key.execution.maybeGet(AddressResolverGroup.class)
```
johnrengelman, line 58: This is the same pattern we do over in `raptack-retrofit2`, but it's ugly and still requires that users explicitly bind or provide an `ExecInitializer` that address `AddressResolverGroup` to forked executions.

ratpack-core/src/main/java/ratpack/server/internal/ServerRegistry.java
```diff
@@ -91,6 +95,11 @@ public static Registry buildBaseRegistry(RatpackServer ratpackServer, Imposition
         .maxContentLength(serverConfig.getMaxContentLength())
       );
 
+      DnsNameResolverBuilder resolverBuilder = new DnsNameResolverBuilder()
+        .eventLoop(execController.getEventLoopGroup().next())
+        .channelType(TransportDetector.getDatagramChannelImpl())
+        .socketChannelType(TransportDetector.getSocketChannelImpl());
```
johnrengelman, line 101: This is essentially the bare minimum for the resolver. 

Should we provide some type of configuration via `ServerConfig` to provide tunings here or just ask the user to provide their own implementation of `AddressResolverGroup` on the Registry?

ratpack-test/src/main/java/ratpack/test/internal/BlockingHttpClient.java
```diff
@@ -46,6 +48,7 @@ public ReceivedResponse request(HttpClient httpClient, URI uri, ExecController e
     AtomicReference<ExecResult<ReceivedResponse>> result = new AtomicReference<>();
 
     execController.fork()
+      .register(r -> r.add(AddressResolverGroup.class, DefaultAddressResolverGroup.INSTANCE))
```
johnrengelman, line 51: This does the necessary registry add so that the `HttpClient` works, but could be removed if we a more holistic approach to this.

ratpack-core/src/main/java/ratpack/server/internal/ServerRegistry.java
```diff
@@ -116,6 +125,7 @@ public static Registry buildBaseRegistry(RatpackServer ratpackServer, Imposition
         .add(Clock.class, Clock.systemDefaultZone())
         .add(RatpackServer.class, ratpackServer)
         .add(ObjectMapper.class, new ObjectMapper())
+        .add(AddressResolverGroup.class, new DnsAddressResolverGroup(resolverBuilder))
```
johnrengelman, line 128: Users can switch back to the default JVM implementation by doing:
```
.add(AddressResolverGroup.class, DefaultAddressResolverGroup.INSTANCE)
```
in the user registry.

ratpack-core/src/main/java/ratpack/server/internal/ServerRegistry.java
amityatagiri: It will be very beneficial if we can configure the dns query timeout at the least as default. 
But, there is already a way to provide few configurations with this change:  https://github.com/netty/netty/commit/4f72cdf2333f682410e23e1eb5d77bc66b186cf4#diff-e7ae8540514a0e3caa415abc27aaee9fe93b7732d318d05a9534686447a162b9L146 .
Other tuning configs would become tricky to expose here just because they are internal to netty and are in plenty.

johnrengelman: I didn't see that previously @amityatagiri, good find.
It does look like the query time is configured based on the the `timeout:` option in `/etc/resolv.conf` (https://github.com/netty/netty/blob/4.1/resolver-dns/src/main/java/io/netty/resolver/dns/UnixResolverDnsServerAddressStreamProvider.java#L334).
So if this is configurable here, what about the benefit of providing a config option inside of Ratpack?
I'd probably lean toward not providing any options here since they are configurable and obtained from the OS by default and ask the user to supply a fully configured `AddressResolverGroup` if they want to tweak things.

ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
johnrengelman: I changed this pattern, so that we looking in the execution and if nothing is there, we get the default netty backed resolver (which defaults its configuration to the system properties in `/etc/resolv.conf`). This means a user will still need to explicitly ensure that `AddressResolverGroup` is registered in any forked executions (including `ParallelBatch`) if they do any customizations.

Also, not sure what the impact of creating/gc'ing those would be....it doesn't look like there's any state really.

One oddity here, is that we are finding the `AddressResolverGroup` from the `Execution`, if you are using Guice and injecting `HttpClient`, the `AddressResolverGroup` doesn't come along with it. It's almost as though `HttpClient` should also inject `AddressResolverGroup`, but the Ratpack `Registry` doesn't do dependency binding.

ratpack-core/src/main/java/ratpack/server/internal/ServerRegistry.java
```diff
@@ -58,6 +62,16 @@
 import static ratpack.util.internal.ProtocolUtil.HTTP_SCHEME;
 
 public abstract class ServerRegistry {
+
+  public static AddressResolverGroup<?> defaultAddressResolver(ExecController controller) {
+    DnsNameResolverBuilder resolverBuilder = new DnsNameResolverBuilder()
+      .eventLoop(controller.getEventLoopGroup().next())
```
johnrengelman, line 68: This does mean that all DNS resolves are bound to 1 event loop...which, since it's NIO _shouldn't_ be a problem (and it is how `reactor-netty` does it too).

ratpack-core/src/main/java/ratpack/server/internal/ServerRegistry.java
amityatagiri: 1.9.x branch is still pointing to netty `4.1.48.Final`. We may want to match it with the master branch or to the latest netty.

johnrengelman: Yes, 
`master` is the 2.x branch right now, that's why it's updated. But we can do that updated in `1.9.x` as well. @amityatagiri can you open another PR that is targeted to `1.9.x` which that dependency change?

ratpack-core/src/main/java/ratpack/http/client/internal/DefaultHttpClient.java
```diff
@@ -88,6 +95,10 @@ private DefaultHttpClient(DefaultHttpClient.Spec spec) {
     } else {
       proxy = null;
     }
+    if (spec.resolver == null) {
+      LOGGER.warn("DNS resolver not specified for HttpClientSpec, defaulting to JVM blocking implementation");
+      spec.resolver = HttpClient.blockingResolver();
```
johnrengelman, line 100: Do the only sane thing we can if someone is creating a new HttpClient and doesn't provide a resolver, which is use the blocking implementation but WARN about it.

We can't default to a non-blocking implementation because we don't have a handle to an `EventLoop` which is needed. We could defer to to the point that we make the request but that would create high GC churn and possible performance impacts as each request would create its own resolver.

ratpack-core/src/main/java/ratpack/http/client/HttpClient.java
```diff
@@ -93,6 +99,46 @@ static HttpClient of(Action<? super HttpClientSpec> action) throws Exception {
...
+  /**
+   * Provides a non-blocking address resolver group implementation.
+   *
+   * @param controller the controller which can provide an event loop for executing resolutions
+   * @param spec provides additional configuration to the resolver
+   * @return a non-blocking resolver
+   * @since 1.9.0
+   */
+  static AddressResolverGroup<?> addressResolver(ExecController controller, Action<? super DnsNameResolverBuilder> spec) throws Exception {
+    DnsNameResolverBuilder resolverBuilder = new DnsNameResolverBuilder()
```
johnrengelman, line 122: The builder automatically loads defaults from `/etc/resolv.conf`.

ratpack-test/src/main/java/ratpack/test/http/internal/DefaultTestHttpClient.java
```diff
@@ -256,6 +256,7 @@ private HttpClient httpClient() {
       .byteBufAllocator(TestByteBufAllocators.LEAKING_UNPOOLED_HEAP)
       .maxContentLength(Integer.MAX_VALUE)
       .poolSize(8)
+      .addressResolver(HttpClient.blockingResolver())
```
johnrengelman, line 259: Since this is the blocking client for tests, use the blocking resolver and avoid the warning.

amityatagiri (approved):
LGTM

ldaley:
@johnrengelman pls check my changes. I haven't observed the full CI response yet, but will do so in the next few days.

johnrengelman:
@ldaley Looks good. I pushed fixes for the broken tests. Biggest piece of that was some HttpClient metrics classes that were not elevated to the public API so we were casting to `DefaultHttpClient` to get them. I've moved those to the public API which requires any `HttpClient` to now implement them.

ldaley:
@johnrengelman I reverted the move to make the stats public. It wasn't necessary, and should be done as its own change if that's what we want to do. 

On that, I'm not sure it _is_ what we want to do. As is stands, we are putting the core client on the hook for maintaining too much information by having it be responsible for providing “stats”. What we should do here instead is allow “event listeners” to be attached that can be notified of various things happening, and they can interpret that into whatever kind of stats they want and with what retention they want.

johnrengelman:
Yeah. That’s a better plan. Something we should address in trunk. 

This looks good to me.

PR #1584, opened 2021-02-12
Add Predicate and Class onError handlers to Operation

timyates:
`Promise` has `onError(Class<Throwable>, ...)` and `onError(Predicate<Throwable>, ...)` methods

These were missing from Operation, so this PR adds them in

I realise you can simply do

`o.promise().onError(E.class, ...).operation()`, but this hides it away

I basically used the promise methods in the operation classes...  I hope this is ok 🤔

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1584)

### DIFF
--- a/ratpack-exec/src/main/java/ratpack/exec/Operation.java
+++ b/ratpack-exec/src/main/java/ratpack/exec/Operation.java
@@ -22,6 +22,7 @@ import ratpack.func.Action;
 import ratpack.func.Block;
 import ratpack.func.Factory;
 import ratpack.func.Function;
+import ratpack.func.Predicate;
 
 import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


ratpack-exec/src/main/java/ratpack/exec/Operation.java
```diff
@@ -86,7 +87,38 @@ static Operation flatten(Factory<Operation> factory) {
...
+  /**
+   * Specifies the action to take if the an error occurs performing the operation that the given predicate applies to.
+   * <p>
+   * If the given action throws an exception, the original exception will be rethrown with the exception thrown
+   * by the action added to the suppressed exceptions list.
+   *
+   * @param predicate the predicate to test against the error
+   * @param errorHandler the action to take if an error occurs
+   * @return An operation for the successful result
+   * @since 1.8.3
```
timyates, line 103: Presumptuous 😜

pio-kol (approved):

ratpack-exec/src/main/java/ratpack/exec/Operation.java
johnrengelman: public API has to be added in a minor release. Luckily....we already have a 1.9.0 in the works.
Can you update this comment and target the PR to the 1.9.x branch?

timyates: Done 😄 

Cheers John 👍

ldaley:
Thanks Tim.

PR #1592, opened 2021-03-09
Require registration of types before they can be used in a session

ldaley:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1592)


ldaley:
@johnrengelman @JLLeitschuh ping

ratpack-session/src/main/java/ratpack/session/AllowedSessionType.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Indicates that the annotated type is always safe to store in a session.
+ * <p>
+ * The default implementation of {@link SessionTypeFilter} provided by {@link SessionModule}
+ * allows any class that carries this annotation to be used in a session.
+ *
+ * @see SessionModule
+ * @since 1.9
+ */
+@Inherited
```
JLLeitschuh, line 30: This is a bad idea. I would not suggest making this an `@Inherited` annotation.

ratpack-session/src/main/java/ratpack/session/internal/AllowAllSessionTypeFilter.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ratpack.session.internal;
+
+import ratpack.session.SessionTypeFilter;
+
+public class AllowAllSessionTypeFilter implements SessionTypeFilter {
```
JLLeitschuh, line 21: This should have a name like "Unsafe" or "Insecure" in the name and also be marked Deprecated.

ratpack-session/src/main/java/ratpack/session/internal/DefaultSessionTypeFilter.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class DefaultSessionTypeFilter implements SessionTypeFilter {
+
+  public static final DefaultSessionTypeFilter INSTANCE = new DefaultSessionTypeFilter();
+  private static final Set<Package> ALLOWED_PACKAGES = Sets.newIdentityHashSet();
+  private static final Set<Class<?>> ALLOWED_CLASSES = Sets.newIdentityHashSet();
+
+  static {
+    ALLOWED_PACKAGES.add(String.class.getPackage()); // java.lang
+    ALLOWED_PACKAGES.add(List.class.getPackage()); // java.util
```
JLLeitschuh, line 37: Will this mean that all subtypes are also supported? If so, this is not a good idea.

ratpack-session/src/main/java/ratpack/session/internal/JavaBuiltinSessionSerializer.java
```diff
@@ -40,4 +41,36 @@
...
+    }
+
+    @Override
+    protected void writeClassDescriptor(ObjectStreamClass desc) throws IOException {
+      typeFilter.assertAllowed(desc.forClass());
+      super.writeClassDescriptor(desc);
+    }
+  }
+
+  private static final class TypeFilteringObjectInputStream extends ObjectInputStream {
```
JLLeitschuh, line 60: Looks like this is a less feature-rich version of the `ValidatingObjectInputStream` from Apache Commons.

I just want to make sure that you're not reinventing the wheel here.

https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/serialization/ValidatingObjectInputStream.java

JLLeitschuh (changes requested):
A few comments. Happy to discuss 1:1 if needed.

ratpack-session/src/main/java/ratpack/session/internal/ClassWithStaticInit.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package ratpack.session.internal;
+
+public class ClassWithStaticInit {
+
+  static {
+    System.out.println("HERE!!!!!!!!!!!!!!!!!!!");
```
johnrengelman, line 22: This seems like a debugging class that shouldn't have been comimtted?

ratpack-session/src/test/groovy/ratpack/session/SessionTypeFilteringSpec.groovy
```diff
@@ -0,0 +1,131 @@
...
+  public <T> void test(Class<T> type, T obj, @Nullable Class<?> failingType = null) {
+    addHandlersFor(type, obj)
+    testFor(failingType)
+  }
+
+  def "allows JDK types"() {
+    expect:
+    test(String, "foo")
+    test(Integer, 1)
+    test(List, [[1, 2] as int[]])
```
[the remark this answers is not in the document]
ldaley: No, you also need to register the component type.

If `Integer` wasn't registered this would fail.

ratpack-session/src/main/java/ratpack/session/internal/ClassWithStaticInit.java
ldaley: What gave it away?

Will remove.

ratpack-session/src/main/java/ratpack/session/internal/JavaBuiltinSessionSerializer.java
ldaley: We don't need any of those features, or the extra dependency.

ratpack-session/src/main/java/ratpack/session/internal/DefaultSessionTypeFilter.java
ldaley: I ended up removing this.

ratpack-session/src/main/java/ratpack/session/internal/AllowAllSessionTypeFilter.java
ldaley: done

ratpack-session/src/main/java/ratpack/session/AllowedSessionType.java
ldaley: the annotation is gone entirely.

It turns out that inspecting class annotations causes the class to be initialized. Theoretically, a class could do something malicious in its static initializer.

ratpack-session/src/main/java/ratpack/session/AllowedAnnotationSessionTypeFilterPlugin.java
```diff
@@ -0,0 +1,25 @@
...
+
+package ratpack.session;
+
+public class AllowedAnnotationSessionTypeFilterPlugin implements SessionTypeFilterPlugin {
+
+  @Override
+  public boolean allow(String className) {
+    return false;
+  }
+}
```
JLLeitschuh, line 25: I'm confused what this is doing. It seems like it just forbids all classes? Some documentation explaining what's going on here would be helpful.

ratpack-session/src/main/java/ratpack/session/internal/DefaultSessionKey.java
```diff
@@ -19,12 +19,16 @@
 import ratpack.api.Nullable;
 import ratpack.session.SessionKey;
 
+import java.util.Objects;
+
 public class DefaultSessionKey<T> implements SessionKey<T> {
```
JLLeitschuh, line 24: I don't have enough context here, but doesn't unintentionally having a `DefaultSessionKey(null, null)` potentially introduce a security risk in itself?

ratpack-session/src/main/java/ratpack/session/AllowedAnnotationSessionTypeFilterPlugin.java
ldaley: Relic. Now removed.

ratpack-session/src/main/java/ratpack/session/internal/DefaultSessionKey.java
ldaley: That state isn't possible. I've moved the check to make that clearer.

I can't see how it would be a risk. There'd be a runtime error before trying to do anything with it.

johnrengelman (approved):

PR #1611, opened 2021-08-09
Update manual for Gradle configuration names that cause errors in Gradle 7 (runtime, etc.)

davidmc24:
`runtime` produces an error `Could not find method runtime() for arguments [org.slf4j:slf4j-simple:1.7.30] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.` in Gradle 7.1.1.

`runtimeOnly` has been present since Gradle 5.0 in November 2018.
If we want to support pre-Gradle 5.0, we could add a comment to note the alternative (as I have done in this PR).

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1611)

### DIFF
--- a/ratpack-manual/src/content/chapters/02-quick-start.md
+++ b/ratpack-manual/src/content/chapters/02-quick-start.md
@@ -81,7 +81,7 @@ repositories {
 }
 
 dependencies {
-  runtime "org.slf4j:slf4j-simple:@slf4j-version@"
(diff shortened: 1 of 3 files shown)


johnrengelman:
Updated in 033780ea05ef397ef237e4f400b51ef69c53cd48

PR #1613, opened 2021-09-29
Configure the main class in RatpackGroovyPlugin without using a method which is deprecated in Gradle 7

erdi:
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1613)


erdi:
I will have to close this due to the failure in `ShadowJarSpec`. It looks like Shadow plugin only switched from using the deprecated `ApplicationPluginConvention` to `JavaApplication` extension in the 7.0 version of the plugin but that version is only compatible with Gradle 7 yet this repo is still on 6.x.

PR #1623, opened 2022-02-02
Fix FileReadingPublisher backpressure

### LINKED ISSUE #1622: FileReadingPublisher doesn't respect back-pressure?
The FileReadingPublisher I get from FileIo.readStream appears not to respect back-pressure under some circumstances.

Specifically: if on the first _n_ invocations of Subscriber.onNext it calls Subscription.request(1) but then stops doing so (e.g. because some buffer is full) then additional _n-1_ items are emitted; this causes the _demand_ in ManagedSubscription to become negative and on the next Subscription.request that _demand_ is set to Long.MAX_VALUE

It happens because recursive calls to request from a downstream onNext do not check the reading mutex.

Test and fix #1623

LeeWorrall:
Fixes #1622

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1623)


johnrengelman:
Applied in b87dc1198

PR #1653, opened 2022-07-01
Add support for HTTP Proxy authentication, HTTPS proxy connections, request-time config overrides, and a TestHttpProxyServer

### LINKED ISSUE #1652: Add support for HTTP Proxy Authentication
Ratpack has support for HTTP proxies, but not for HTTP proxies which require authentication. Netty has support for proxy authentication, so this should be as easy as adding username and password properties to ProxySpec and passing them to Netty's HttpProxyHandler if they are present.

Basic authentication is the primary method of authentication used by HTTP proxies (and this is what Netty supports). Other forms of authentication do not need to be supported.

jthurne:
Netty supports HTTP Proxy authentication, but Ratpack didn't give users a way
to specify the username and password to use when authenticating with an HTTP
Proxy.

Also adds support for connecting to an HTTP Proxy using SSL, and for overriding the proxy settings at request-time (on the RequestSpec).

Also adds a TestHttpProxyServer useful for testing proxy code.

#1652

---
This change is [Reviewable](https://reviewable.io/reviews/ratpack/ratpack/1653)

### DIFF
--- a/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
@@ -44,7 +44,7 @@ public interface Proxy {
   /**
    * The username to use when connecting to the proxy.
    *
-   * @return The username to use when connecting to the HTTP proxy, or null if a username should not be used when connecting to the HTTP proxy.
+   * @return The username to use when connecting to the HTTP proxy, or null if authentication should not be used when connecting to the HTTP proxy.
    */
   @Nullable
   String getUsername();
@@ -52,7 +52,7 @@ public interface Proxy {
   /**
    * The password to use when connecting to the proxy.
    *
-   * @return Either the password to use when connecting to the HTTP proxy, or null if a password should not be used when connecting to the HTTP proxy.
+   * @return Either the password to use when connecting to the HTTP proxy, or null if authentication should not be used when connecting to the HTTP proxy.
    */
   @Nullable
   String getPassword();
--- a/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
@@ -16,6 +16,8 @@
 
 package ratpack.http.client;
 
+import ratpack.api.Nullable;
+
 import java.util.Collection;
 
 /**
@@ -38,6 +40,22 @@ public interface Proxy {
    * @return The port on the proxy where proxied requests will be sent.
    */
   int getPort();
+
+  /**
+   * The username to use when connecting to the proxy.
+   *
+   * @return The username to use when connecting to the HTTP proxy, or null if a username should not be used when connecting to the HTTP proxy.
+   */
+  @Nullable
+  String getUsername();
+
+  /**
+   * The password to use when connecting to the proxy.
+   *
+   * @return Either the password to use when connecting to the HTTP proxy, or null if a password should not be used when connecting to the HTTP proxy.
+   */
+  @Nullable
+  String getPassword();
 
   /**
    * A collection of patterns which if any or matched, the outgoing request will bypass the HTTP proxy.
--- a/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
@@ -31,7 +31,8 @@ public interface Proxy {
     HTTP, HTTPS
   }
 
-  /** The protocol that should be used when connecting to the proxy server itself.
+  /**
+   * The protocol that should be used when connecting to the proxy server itself.
    *
    * @return the protocol to use to connect to the proxy server
    */
--- a/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
@@ -27,6 +27,16 @@ import java.util.Collection;
  */
 public interface Proxy {
 
+  enum Protocol {
+    HTTP, HTTPS
+  }
+
+  /** The protocol that should be used when connecting to the proxy server itself.
+   *
+   * @return the protocol to use to connect to the proxy server
+   */
+  Protocol getProtocol();
+
   /**
    * The host that proxied requests will be sent.
    *
--- a/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
@@ -42,20 +42,12 @@ public interface Proxy {
   int getPort();
 
   /**
-   * The username to use when connecting to the proxy.
+   * The credentials to use when connecting to the proxy.
    *
-   * @return The username to use when connecting to the HTTP proxy, or null if authentication should not be used when connecting to the HTTP proxy.
+   * @return The credentials to use when connecting to the HTTP proxy, or null if authentication should not be used when connecting to the HTTP proxy.
    */
   @Nullable
-  String getUsername();
-
-  /**
-   * The password to use when connecting to the proxy.
-   *
-   * @return Either the password to use when connecting to the HTTP proxy, or null if authentication should not be used when connecting to the HTTP proxy.
-   */
-  @Nullable
-  String getPassword();
+  ProxyCredentials getCredentials();
 
   /**
    * A collection of patterns which if any or matched, the outgoing request will bypass the HTTP proxy.
--- a/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/Proxy.java
@@ -27,7 +27,7 @@ import java.util.Collection;
  */
 public interface Proxy {
 
-  enum Protocol {
+  enum ProxyProtocol {
     HTTP, HTTPS
   }
 
@@ -36,7 +36,7 @@ public interface Proxy {
    *
    * @return the protocol to use to connect to the proxy server
    */
-  Protocol getProtocol();
+  ProxyProtocol getProtocol();
 
   /**
    * The host that proxied requests will be sent.
--- a/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultProxy.java
+++ b/ratpack-core/src/main/java/ratpack/http/client/internal/DefaultProxy.java
@@ -17,6 +17,7 @@
 package ratpack.http.client.internal;
 
 import ratpack.api.Nullable;
+import ratpack.http.client.Proxy;
 import ratpack.http.client.ProxyCredentials;
 import ratpack.http.client.ProxySpec;
 
@@ -26,17 +27,24 @@ import java.util.Objects;
 
 public class DefaultProxy implements ProxyInternal {
 
+  private final Protocol protocol;
   private final String host;
   private final int port;
   private final Collection<String> nonProxyHosts;
 
   private final ProxyCredentials credentials;
 
-  public DefaultProxy(String host, int port, Collection<String> nonProxyHosts, @Nullable ProxyCredentials credentials) {
+  public DefaultProxy(Protocol protocol, String host, int port, Collection<String> nonProxyHosts, @Nullable ProxyCredentials credentials) {
+    this.protocol = protocol;
     this.host = host;
     this.port = port;
     this.nonProxyHosts = nonProxyHosts;
     this.credentials = credentials;
+  }
+
+  @Override
+  public Protocol getProtocol() {
+    return protocol;
   }
 
   @Override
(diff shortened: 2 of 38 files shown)


ratpack-core/src/main/java/ratpack/http/client/Proxy.java
```diff
@@ -39,6 +41,22 @@ public interface Proxy {
    */
   int getPort();
 
+  /**
+   * The username to use when connecting to the proxy.
+   *
+   * @return The username to use when connecting to the HTTP proxy, or null if a username should not be used when connecting to the HTTP proxy.
+   */
+  @Nullable
+  String getUsername();
```
jthurne, line 50: I considered using `Optional<String>`, but it looks like nullable fields and null-checks are more common for user-provided values in Ratpack's public API.

ratpack-core/src/test/groovy/ratpack/http/client/HttpForwardProxySpec.groovy
```diff
@@ -86,4 +93,78 @@ class HttpForwardProxySpec extends BaseHttpClientSpec {
...
+    AtomicBoolean proxied = new AtomicBoolean(false)
+    def username = "testUser"
+    def password = "testPassword"
+
+    proxyApp {
+      all {
+        def authHeader = request.headers.get(PROXY_AUTHORIZATION)
+        if (authHeader == null) {
+          if (authenticated.get()) {
+            // The client was already authenticated. In real life, we would have validated the client's session.
```
jthurne, line 108: When developing this test, I discovered that Netty sends to requests to the HTTP Proxy: the first contains the Proxy-Authorization header, while the second does not.

Rather than adding session handling and such, it was easier in this test to just track when the first request was authenticated and accepted.

davidmc24 (approved):
LGTM

davidmc24:
@jthurne I think an entry in `release-notes.md` would be a good idea.

jthurne:
> @jthurne I think an entry in release-notes.md would be a good idea.

Done.

ldaley (changes requested):
Please introduce a `ProxyCredentials` interface with `username` and `password` getters, and change `ProxySpec` to have a `credentials(String username, String password)` method. 

Effectively, the user/pass needs to be one object so that we can more reasonably enforce the nullability contract.

jthurne:
> Please introduce a `ProxyCredentials` interface with `username` and `password` getters, and change `ProxySpec` to have a `credentials(String username, String password)` method.
> 
> Effectively, the user/pass needs to be one object so that we can more reasonably enforce the nullability contract.

👍 Will do.

Do we need to provide a way for users to "unset" or "disable" authentication on a ProxySpec (if it has already been set)?

jthurne:
> Please introduce a ProxyCredentials interface

Done.

ratpack-core/src/main/java/ratpack/http/client/internal/HttpChannelKey.java
```diff
@@ -17,25 +17,33 @@
...
 import java.time.Duration;
+import java.util.Objects;
 
 final class HttpChannelKey {
 
   final boolean ssl;
   final int port;
   final String host;
 
+  final ProxyInternal proxy;
```
jthurne, line 34: I added ProxyInternal to the HttpChannelKey because Proxy settings influence how the client connects to the destination. This was partly to make them accessible to the ChannelPoolHandler.channelCreated where the proxy Netty handler is created.

I tried moving the handler creation to RequestActionSupport, but by the time RequestActionSupport gets the channel, it's too late to add the proxy handler(s).

A positive consequence of this is that proxy connections, including their tunnels, are pooled. Thus, when using proxies, the client does not always have to establish a new tunnel to the destination.

The proxy settings are only added to the HttpChannelKey if there are proxy settings. Thus, is no additional memory overhead in the common case where proxies are not used.

ratpack-test/src/main/java/ratpack/test/http/proxy/TestHttpProxyServer.java
```diff
@@ -0,0 +1,160 @@
...
+import java.net.URI;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import static ratpack.test.http.proxy.internal.ProxyHandlerNames.HTTP_CODEC_HANDLER;
+
+// TODO Add class-level and method-level Javadoc
+public class TestHttpProxyServer implements AutoCloseable {
```
jthurne, line 37: There wasn't a suitable embeddable, Java-based test HTTP proxy available from the open source community, so I decided to write one (Netty has one in their internal test code, but Netty doesn't distribute their internal test code).

The Ratpack HttpClient (and Netty) currently only supports using the CONNECT method, so the TestProxyServer only supports CONNECT as well.

TestHttpServer is written using just Netty (and not Ratpack). The CONNECT method of connecting to an HTTP Proxy requires establishing a low-level TCP tunnel, which is hard to accomplish with Ratpack.

I haven't yet gotten a chance to write the Javadoc for this class, and how it is configured is a bit messy right now.

I added this to Ratpack's public test library so that applications built on Ratpack can use it.

ratpack-core/src/main/java/ratpack/http/client/internal/ProxyChannelPoolHandler.java
```diff
@@ -0,0 +1,67 @@
...
+import io.netty.handler.proxy.HttpProxyHandler;
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslContextBuilder;
+import io.netty.handler.ssl.SslHandler;
+
+import javax.net.ssl.SSLException;
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
+
+public class ProxyChannelPoolHandler extends AbstractChannelPoolHandler {
```
jthurne, line 30: Inheritance appears to be how shared logic is handled with the ChannelPoolHanders, so this commit creates a ProxyChanelPoolHandler that the NoopXXXChannelPoolHanders extend.

ratpack-core/src/main/java/ratpack/http/client/internal/HttpChannelKey.java
jthurne, line 34: The ProxyInternal has an SslContext property that is a copy of the SslContext set on a RequestSpec.  This SslContext property is not 

However, the SslContext is _not_ used in equality checks. SslContext does not implement equals and hashcode, and writing a Comparator to do so or similar would be complex. Instead, Proxy.protocol is the only new property that influences equality.

This means that some request-level changes to the SslContext will _not_ result in a new connection being established with the proxy. There are some rare use cases that this might create problems for (such as if SSL client authentication is in use, and different client certificates result in different proxy behavior), but these are consistent with Ratpack's HttpCient SSL support in general (direct connections also have this same problem).

The SslContext is only saved on a ProxyInternal (and therefore HttpChannelKey) if the user
has configured an HTTP Proxy and has set the Proxy protocol to HTTPS. This eliminates the memory increase for the common case where proxies are not used, or when SSL is not used to communicate to the Proxy.

ratpack-core/src/main/java/ratpack/http/client/internal/DefaultProxy.java
```diff
@@ -96,32 +126,91 @@ private boolean matchNonProxyHost(String targetHost, String nonProxyHost) {
...
+    }
 
     @Override
-    public ProxySpec host(String host) {
+    public Builder protocol(Protocol protocol) {
+      this.protocol = protocol;
+      return this;
+    }
+    @Override
+    public Builder host(String host) {
```
davidmc24, line 184:
```suggestion
    public Builder protocol(Protocol protocol) {
      this.protocol = protocol;
      return this;
    }

    @Override
    public Builder host(String host) {
```

ratpack-test/src/main/java/ratpack/test/http/proxy/TestHttpProxyServer.java
```diff
@@ -0,0 +1,175 @@
...
+import java.net.InetSocketAddress;
+import java.net.URI;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import static ratpack.test.http.proxy.internal.ProxyHandlerNames.HTTP_CODEC_HANDLER;
+
+// TODO Add class-level and method-level Javadoc
```
davidmc24, line 51: Pending TODO?

ratpack-test/src/main/java/ratpack/test/http/proxy/TestHttpProxyServer.java
```diff
@@ -0,0 +1,175 @@
...
+public class TestHttpProxyServer implements AutoCloseable {
+
+  public static final Logger LOGGER = LoggerFactory.getLogger(TestHttpProxyServer.class);
+
+  // TODO can we reuse Ratpack's test-scoped ExecutionContext in some way to avoid creating another set of EventLoopGroups?
```
davidmc24, line 56: Pending TODO?

ratpack-test/src/main/java/ratpack/test/http/proxy/TestHttpProxyServer.java
```diff
@@ -0,0 +1,175 @@
...
+  private ConcurrentMap<Authority, Integer> numTunnelsByDestination = new ConcurrentHashMap<>();
+
+  /**
+   * Starts the HTTP Proxy server without requiring authentication.
+   */
+  public void start() {
+    this.start(false);
+  }
+
+  // TODO Refactor to use a TestProxyServerSpec instead of taking proxy server configuration parameters directly in the start method
```
davidmc24, line 73: Pending TODO?

ratpack-test/src/main/java/ratpack/test/http/proxy/TestHttpProxyServer.java
```diff
@@ -0,0 +1,175 @@
...
+  public void start(boolean requireSsl) {
+    this.start(null, requireSsl);
+  }
+
+  /** Starts the HTTP proxy server with proxy authentication enabled.
+   *
+   * @param username the username clients must provide to authenticate
+   * @param password the password clients must provide to authenticate
+   */
```
davidmc24, line 82:
```suggestion
  /**
   * Starts the HTTP proxy server with proxy authentication enabled.
   *
   * @param username the username clients must provide to authenticate
   * @param password the password clients must provide to authenticate
   */
```

ratpack-test/src/main/java/ratpack/test/http/proxy/TestHttpProxyServer.java
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  public void start(String username, String password, boolean requireSsl) {
+    this.start(new DefaultProxyCredentials(username, password), requireSsl);
+  }
+
+  /** Starts the HTTP proxy server. If credentials are not null, then basic proxy authentication is required.
+   *
+   * @param credentials the username/password to require for proxy authentication, or null if no authentication is required
+   */
```
davidmc24, line 94:
```suggestion
  /**
   * Starts the HTTP proxy server. If credentials are not null, then basic proxy authentication is required.
   *
   * @param credentials the username/password to require for proxy authentication, or null if no authentication is required
   */
```

davidmc24 (changes requested):

ratpack-core/src/main/java/ratpack/http/client/Proxy.java
```diff
@@ -25,6 +27,16 @@
  */
 public interface Proxy {
 
+  enum Protocol {
+    HTTP, HTTPS
+  }
```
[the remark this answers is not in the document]
jthurne: Renamed to `ProxyProtocol`

davidmc24 (approved):

ratpack-core/src/main/java/ratpack/http/client/Proxy.java
ldaley: ACK.

ratpack-core/src/test/groovy/ratpack/http/client/HttpForwardProxySpec.groovy
ldaley: ACK.

ratpack-test/src/main/java/ratpack/test/http/proxy/TestHttpProxyServer.java
ldaley: No problem now that it's internal.

ratpack-test/src/main/java/ratpack/test/http/proxy/TestHttpProxyServer.java
ldaley: No problem now that it's internal.

ratpack-test/src/main/java/ratpack/test/http/proxy/TestHttpProxyServer.java
ldaley: No problem now that it's internal.

ratpack-test/src/main/java/ratpack/test/http/proxy/TestHttpProxyServer.java
ldaley: No problem now that it's internal.

ldaley:
Thank you @jthurne

ldaley (approved):

PR #1717, opened 2024-08-30
Improve `@NonBlocking` usage

leonard84:
Add `TYPE` as target to bring it in line with
`org.jetbrains.annotations.NonBlocking`.
Annotate `Promise` and `Operation` as `@NonBlocking` to facilitate static checking of calling blocking code in a non-blocking context.

https://www.jetbrains.com/help/inspectopedia/BlockingMethodInNonBlockingContext.html


ldaley:
Superseded by https://github.com/ratpack/ratpack/pull/1720.

PR #1719, opened 2024-10-02
Change redirect encoding mechanism to not decode parameters

gcoopercos:
This stemmed from an internal issue where the redirect url was getting decoded, and thus changing from the original. 

It looked like there were test around the query parameters, but not so much for path parameters.  The change make it such that the path parameters do not get decoded if they are encoded.  Thus, something like this: `some-data/endpoint/nslowehwwgbgm/target/%2F%2F%3AHelloWorldTest` would end up getting mapped to an endpoint that matches `../target/`.

FWIW, this is our internal issue that was a motivator for this: https://github.com/gradle/dv/issues/40337

### DIFF
--- a/ratpack-core/src/main/java/ratpack/core/http/client/internal/RequestActionSupport.java
+++ b/ratpack-core/src/main/java/ratpack/core/http/client/internal/RequestActionSupport.java
@@ -618,7 +618,7 @@ abstract class RequestActionSupport<T> implements Upstream<T> {
         String rawQuery = redirectLocationUri.getRawQuery() != null
           ? "?" + redirectLocationUri.getRawQuery() : "";
         return new URI(new URI(requestUri.getScheme(),
-          requestUri.getRawUserInfo(),
+          requestUri.getUserInfo(),
           requestUri.getHost(),
           requestUri.getPort(),
           null,
--- a/ratpack-core/src/main/java/ratpack/core/http/client/internal/RequestActionSupport.java
+++ b/ratpack-core/src/main/java/ratpack/core/http/client/internal/RequestActionSupport.java
@@ -608,17 +608,17 @@ abstract class RequestActionSupport<T> implements Upstream<T> {
         return URI.create(requestUri.getScheme() + ":" + redirectLocation);
       } else {
 
-        String path = redirectLocationUri.getPath();
+        String path = redirectLocationUri.getRawPath();
         if (!path.startsWith("/")) { // absolute path
-          path = getParentPath(requestUri.getPath()) + path;
+          path = getParentPath(requestUri.getRawPath()) + path;
         }
         return new URI(
           requestUri.getScheme(),
-          requestUri.getUserInfo(),
+          requestUri.getRawUserInfo(),
           requestUri.getHost(),
           requestUri.getPort(),
           path,
-          redirectLocationUri.getQuery(),
+          redirectLocationUri.getRawQuery(),
           null
         );
       }
--- a/ratpack-core/src/main/java/ratpack/core/http/client/internal/RequestActionSupport.java
+++ b/ratpack-core/src/main/java/ratpack/core/http/client/internal/RequestActionSupport.java
@@ -612,15 +612,18 @@ abstract class RequestActionSupport<T> implements Upstream<T> {
         if (!path.startsWith("/")) { // absolute path
           path = getParentPath(requestUri.getRawPath()) + path;
         }
-        return new URI(
-          requestUri.getScheme(),
+
+        // By default, the URI encodes the path given to it. We don't want that here. But, we can let the URI
+        // do most of the work and then just append the raw path and query afterwords
+        String rawQuery = redirectLocationUri.getRawQuery() != null
+          ? "?" + redirectLocationUri.getRawQuery() : "";
+        return new URI(new URI(requestUri.getScheme(),
           requestUri.getRawUserInfo(),
           requestUri.getHost(),
           requestUri.getPort(),
-          path,
-          redirectLocationUri.getRawQuery(),
-          null
-        );
+          null,
+          null,
+          null) + path + rawQuery);
       }
     }
   }
--- a/ratpack-core/src/test/groovy/ratpack/core/http/client/HttpClientRedirectionSpec.groovy
+++ b/ratpack-core/src/test/groovy/ratpack/core/http/client/HttpClientRedirectionSpec.groovy
@@ -182,11 +182,13 @@ class HttpClientRedirectionSpec extends BaseHttpClientSpec {
     }
 
     then:
-    text == "http://ratpack.io=10"
+    text == "${URLEncoder.encode("http://ratpack.io=10", "UTF-8")}"
 
     where:
     pooled << [true, false]
   }
+
+
 
   def "can follow a relative redirect get request"() {
     given:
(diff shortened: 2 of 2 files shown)


ratpack-core/src/test/groovy/ratpack/core/http/client/HttpClientRedirectionSpec.groovy
```diff
@@ -182,12 +182,14 @@ class HttpClientRedirectionSpec extends BaseHttpClientSpec {
     }
 
     then:
-    text == "http://ratpack.io=10"
+    text == "${URLEncoder.encode("http://ratpack.io=10", "UTF-8")}"
```
davidmc24: This doesn't look right to me.  A query param with a single level of encoding, when accessed through `request.queryParams`, should be decoded.  This test makes it look like that has been broken.

davidmc24: We may need more test coverage to determine if other scenarios are working as intended or not.

gcoopercos: Agreed, definitely will need more test coverage.  I'll put one in that mimics/illustrates the Develocity issue more clearly.

gcoopercos: I backed the assertion out as it appears to be desirable for query parameters to be converted. I did, however, add some additional tests, along with some altered functionality.

gcoopercos:
The heart of this fix is an attempt to support redirects on locations with encoded path parameters. I added a couple tests that illustrate this.

ratpack-core/src/test/groovy/ratpack/core/http/client/HttpClientRedirectionSpec.groovy
```diff
@@ -153,6 +153,41 @@ class HttpClientRedirectionSpec extends BaseHttpClientSpec {
     pooled << [true, false]
   }
 
+  def "can follow a relative redirect get request with non-encoded query parameters"() {
```
davidmc24, line 160: The non-encoded query parameter test cases seem questionable to me.  It's unclear to me what the behavior should be as per the spec.

ratpack-core/src/test/groovy/ratpack/core/http/client/HttpClientRedirectionSpec.groovy
```diff
@@ -188,6 +223,78 @@ class HttpClientRedirectionSpec extends BaseHttpClientSpec {
     pooled << [true, false]
   }
 
+  def "can follow a relative redirect get request with encoded path parameters"() {
```
davidmc24, line 230: For this test case, I think there are two useful values to assert on: the path (as currently done) and the path token (`render context.pathTokens.val`).  The former should be preserved encoded, while the latter should IMO be decoded.  From a quick look, I think this behavior is worked after your changes.

ratpack-core/src/test/groovy/ratpack/core/http/client/HttpClientRedirectionSpec.groovy
```diff
@@ -188,6 +223,78 @@ class HttpClientRedirectionSpec extends BaseHttpClientSpec {
...
+
+    then:
+    text == "scan-data/bazel/nslowehwwgbgm/target/%2F%2F%3AHelloWorldTest"
+
+    where:
+    pooled << [true, false]
+  }
+
+
+  def "can follow a relative redirect get request without encoded path parameters"() {
```
davidmc24, line 269: Asserting on the path token might be valuable here as well.

ratpack-core/src/main/java/ratpack/core/http/client/internal/RequestActionSupport.java
```diff
@@ -608,19 +608,22 @@ private static URI absolutizeRedirect(URI requestUri, String redirectLocation) t
...
-        return new URI(
-          requestUri.getScheme(),
-          requestUri.getUserInfo(),
+
+        // By default, the URI encodes the path given to it. We don't want that here. But, we can let the URI
+        // do most of the work and then just append the raw path and query afterwords
+        String rawQuery = redirectLocationUri.getRawQuery() != null
+          ? "?" + redirectLocationUri.getRawQuery() : "";
+        return new URI(new URI(requestUri.getScheme(),
+          requestUri.getRawUserInfo(),
```
davidmc24: If we're going to change the userinfo behavior, we should have tests for it.  I'd be fine with leaving it unchanged as well.

davidmc24 (changes requested):

ratpack-core/src/main/java/ratpack/core/http/client/internal/RequestActionSupport.java
gcoopercos: I reverted that. It's not really in the spirit of what is being addressed here.

ratpack-core/src/test/groovy/ratpack/core/http/client/HttpClientRedirectionSpec.groovy
gcoopercos: Hmm... Yes, I agree.  I just put that test in to capture the current behavior, which is the same before and after the change here.

Like you indicate, the spec really only speaks of the location header being a URL, and doesn't mention much about the encoded/decoded nature of this value:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections

The current ratpack implementation for query parameters does decode them. This change should only be impacting path parameters.

So... As a test that indicates that this behavior is the "contracted, expected result", it's a bit ambiguous, and probably not so great. But, as a regression test indicating that nothing has changed if future work is done in this are, it has value. 

I kept the test, but put in a comment.  Would that work for you?   I can alter wording if it's not clear as well.

gcoopercos: Hmm... Yes, I agree.  I just put that test in to capture the current behavior, which is the same before and after the change here.

Like you indicate, the spec really only speaks of the location header being a URL, and doesn't mention much about the encoded/decoded nature of this value:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections

The current ratpack implementation for query parameters does decode them. This change should only be impacting path parameters.

So... As a test that indicates that this behavior is the "contracted, expected result", it's a bit ambiguous, and probably not so great. But, as a regression test indicating that nothing has changed if future work is done in this are, it has value. 

I kept the test, but put in a comment.  Would that work for you?   I can alter wording if it's not clear as well.

ratpack-core/src/test/groovy/ratpack/core/http/client/HttpClientRedirectionSpec.groovy
gcoopercos: Added.

ratpack-core/src/test/groovy/ratpack/core/http/client/HttpClientRedirectionSpec.groovy
gcoopercos: Added.

ratpack-core/src/test/groovy/ratpack/core/http/client/HttpClientRedirectionSpec.groovy
davidmc24: Yes, that approach is fine with me.

davidmc24 (approved):

